PR #1, opened 2019-06-07
couple small cleanups

mgeplf:
* use utils.load_json when possible
* check against collections.Mappings
* have pandas index match for edges tests

### DIFF
--- a/tests/test_edges.py
+++ b/tests/test_edges.py
@@ -19,6 +19,11 @@ TEST_DIR = os.path.dirname(os.path.abspath(__file__))
 TEST_DATA_DIR = os.path.join(TEST_DIR, "data")
 
 
+def index(values):
+    '''have pandas index types match'''
+    return np.array(values, dtype=np.uint64)
+
+
 def test_get_population_name_duplicate():
     storage = Mock()
     storage.population_names = ['a', 'b']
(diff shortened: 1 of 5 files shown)


tests/test_edges.py
```diff
@@ -120,15 +125,15 @@ def test_properties_1(self):
                 (0, 1, 88.1862, 1111.),
             ],
             columns=properties,
-            index=edge_ids
+            index=index(edge_ids)
```
mgeplf: when I ran the tests under py27/py35 and `pandas==0.24.2`, I was getting:

```
l = UInt64Index([1], dtype='uint64'), r = Int64Index([1], dtype='int64'), obj = 'DataFrame.index'

    def _check_types(l, r, obj='Index'):
        if exact:
>           assert_class_equal(l, r, exact=exact, obj=obj)
E           AssertionError: DataFrame.index are different
E
E           DataFrame.index classes are not equivalent
E           [left]:  UInt64Index([1], dtype='uint64')
E           [right]: Int64Index([1], dtype='int64')

.tox/py35/lib/python3.5/site-packages/pandas/util/testing.py:797: AssertionError

```

arsenius7: should we instead instruct `pandas.util.testing` to ignore `dtype` difference, if there is such an option?

mgeplf: yeah, I can use `check_index_type=False` I think, but I think it's actually good that it's unsigned, so worth checking

tests/test_edges.py
```diff
@@ -19,6 +19,11 @@
 TEST_DATA_DIR = os.path.join(TEST_DIR, "data")
 
 
+def index(values):
```
arsenius7: `as_uint64`?

PR #7, opened 2019-06-14
Require 100% unit test coverage

arsenius7:
* also, bumped 'libsonata' dependency

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - pip install codecov
 
 script:
-  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then tox -e lint,coverage; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then tox -e py36,lint; fi
   - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then tox -e py27; fi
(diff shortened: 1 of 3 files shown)


arsenius7:
If we enable such a policy, `codecov` becomes just a fancy way to show a badge in README, as far as I can tell.
Please let me know if you believe we should leave only one thing (`codecov` integration vs `--cov-fail-under=100`).

setup.py
```diff
@@ -26,7 +26,7 @@
         'cached_property>=1.0',
         'functools32;python_version<"3.2"',
         'h5py>=2.2',
-        'libsonata>=0.0.2',
+        'libsonata>=0.0.3',
```
arsenius7: `uint64` dtype for synapse IDs comes from `libsonata>=0.0.3`; if we use previous `libsonata` versions, the recently fixed tests will be broken.

.travis.yml
```diff
@@ -26,7 +26,7 @@ deploy:
   on:
     tags: true
     branch: master
-    condition: "$TRAVIS_PYTHON_VERSION == 3.6"
+    skip_existing: true
```
arsenius7: @wizmer has pointed out supposedly more proper way to limit PyPI upload to one environment

arsenius7:
> Please let me know if you believe we should leave only one thing (codecov integration vs --cov-fail-under=100).

I guess I'll just submit it the way it is; and let you and Thomas remove either part later, if deemed necessary.

PR #8, opened 2019-06-14
Expose node dynamics params

arsenius7:
See also `dynamics_params` group description at:
https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#nodes---optional-reserved-attributes


tests/test_nodes.py
```diff
@@ -19,13 +19,13 @@
...
 
 
-def test_get_population_name_duplicate():
-    storage = Mock()
-    storage.population_names = ['a', 'b']
-    with patch(test_module.__name__ + '.libsonata.NodeStorage') as NodeStorage:
-        NodeStorage.return_value = storage
-        with pytest.raises(BlueSnapError):
-            test_module._get_population_name(mock.ANY)
+# def test_get_population_name_duplicate():
```
arsenius7: intentionally decreased coverage rate to see the result in CI ^^

bluesnap/nodes.py
```diff
@@ -54,8 +54,10 @@ def _load_population(h5_filepath, csv_filepath, population):
     result = pd.DataFrame(index=np.arange(node_count))
 
     _all = libsonata.Selection([(0, node_count)])
-    for prop in sorted(nodes.attribute_names):
-        result[prop] = nodes.get_attribute(prop, _all)
+    for attr in sorted(nodes.attribute_names):
+        result[attr] = nodes.get_attribute(attr, _all)
+    for attr in sorted(nodes.dynamics_attribute_names):
```
arsenius7: Alternatively, we can let this SONATA idiosyncrasy leak and expose attributes from `dynamics_parameter` in a separate `NodePopulation` method.

tomdele: I am not sure about this. As a user I don t think I would like to learn which attributes are dynamic or not.

tests/test_nodes.py
```diff
@@ -103,6 +103,7 @@ def test_basic(self):
         assert(
             sorted(self.test_obj.property_names) ==
             [
+                Cell.HOLDING_CURRENT,
```
tomdele: no THRESHOLD_CURRENT here ?

arsenius7: These are optional attributes; I have put only `holding_current` attribute to `dynamics_params` in the test data.

tomdele: ok make sense.

PR #9, opened 2019-06-14
Fixed sdist

arsenius7:


tox.ini
```diff
@@ -12,7 +12,7 @@ deps =
     nose
     pytest
     pytest-cov
-commands = py.test tests --cov={[base]name} --cov-fail-under=100
+commands = py.test tests --cov={envsitepackagesdir}/{[base]name} --cov-fail-under=100
```
arsenius7: without `{envsitepackagesdir}`, it does not collect coverage info if there is no `__init__py` in `tests` (and that one came in conflict with something else I'm not eager to investigate, and was excessive anyway).

setup.py
```diff
@@ -18,6 +18,21 @@
...
+from setuptools.command.egg_info import egg_info
+
+
+# setuptools_scm forcibly includes all files under version control into the sdist
+# See https://github.com/pypa/setuptools_scm/issues/190
+# Workaround taken from:
+# https://github.com/raiden-network/raiden/commit/3fb837e8b6e6343f65d99055459cb440e1a938ff
+class EggInfo(egg_info):
+    def __init__(self, *args, **kwargs):
+        egg_info.__init__(self, *args, **kwargs)
```
arsenius7: old-style `super` for Python 2 compatibility

PR #35, opened 2020-03-02
Add Standalone objects for node and edges

tomdele:
This appears to be very handy when someone wants to explore a standalone
file without a circuit config or the proper link to the node files for edges.


tomdele:
Hi guys, 
Should we forget about this ?

PR #39, opened 2020-03-16
Plots

tomdele:
Plot impl in snap


bluepysnap/spike_report.py
```diff
@@ -99,37 +99,53 @@ def get(self, group=None, t_start=None, t_stop=None):
             t_stop (float): Include only spikes occurring before this time.
 
         Returns:
-            pandas.Series/np.array:
-                If single node ID is passed as ``group`` returns a np.array of spiking times.
```
lguyot, line 102: I don't know who was using this function before, so I am just asking you to confirm that there is no code which passes a single node ID anymore.

tomdele: This is not used yet. So this is fine to change it now.

tomdele:
I still need to check stuffs and correct bugs. But the API is more or less fixed I think.

tomdele:
@jdcourcol do you have more comments (except for the group thing) or we can review ?

tomdele:
example : 
```
from bluepysnap import Simulation
from bluepysnap.bbp import Cell
import matplotlib.pyplot as plt
import numpy as np

sim = Simulation("simulation_config.json")
spike_report = sim.spikes

# query examples
spike_report.filter(group=[1, 14944])
spike_report.filter(group={"node_id": [1, 14944, 18000]})
spike_report.filter(group={"population": "population_name"})
spike_report.filter(group={Cell.ETYPE:: "SLM_PPA"})

filtered = spike_report.filter(group={Cell.ETYPE: ['cAC', 'cNAC']}, t_start=50, t_stop=800)
filtered.raster()
filtered.raster(y_axis=Cell.MTYPE)
filtered.firing_rate_histogram()
filtered.isi(binsize=15)
anim, ax = filtered.firing_animation()
ax = sim.reports["soma"].trace(group=np.arange(200), plot_type="all")
plt.show()

# with multiple plots and axes
fig = plt.gcf()
fig.suptitle("Raster plots for hippocampus_neurons")
ax1 = fig.add_subplot(311, sharex=None)
ax1.set_ylabel("mtype")
ax1 = filtered.raster(y_axis=Cell.MTYPE, ax=ax1)
ax2 = fig.add_subplot(312, sharex=ax1)
ax2.set_ylabel("X")
ax2 = filtered.raster(y_axis=Cell.X, ax=ax2)
ax3 = fig.add_subplot(313, sharex=ax1)
ax3.set_ylabel("Y")
ax3 = filtered.raster(y_axis=Cell.Y, ax=ax3)

plt.setp(ax1.get_xticklabels(), visible=False)
plt.setp(ax2.get_xticklabels(), visible=False)
plt.setp(ax3.get_xticklabels(), visible=True)
fig.subplots_adjust(hspace=0)
plt.show()

```

jdcourcol:
> @jdcourcol do you have more comments (except for the group thing) or can we review ?

I am good. 'group' should be renamed to 'predicate' or something along this line.

tomdele:
Thanks a lot @jdcourcol for all the feedback. 
@mgeplf, @haleepfl , @asanin-epfl , @lguyot you can review.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+def spike_raster(filtered_report, y_axis=None, ax=None):  # pragma: no cover
+    """Spike raster plot.
+
+    Shows a global overview of the circuit's firing nodes. The y axis can project either the
+    node_ids or any properties present in the different node populations.
+
+    Args:
+        y_axis (None/str): The property to display on the y axis. None is node_ids.
```
haleepfl, line 96: Wouldn't it be better to make the default `y_axis="node_ids"`? Then the user doesn't have to understand what `None` means in this particular context.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,363 @@
...
+    return plt
+
+
+def spikes_firing_rate_histogram(filtered_report, time_binsize=None, ax=None):  # pragma: no cover
+    """Spike firing rate histogram.
+
+    This plot shows the number of nodes firing during a range of time.
+
+    Args:
+        time_binsize(int): bin size (milliseconds)
```
haleepfl, line 48: What does the default of `None` mean? Include `None` in the type. Is a float ever acceptable?

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,363 @@
...
+
+def spikes_isi(filtered_report, use_frequency=False, binsize=None, ax=None):  # pragma: no cover
+    # pylint: disable=too-many-locals
+    """Interspike interval histogram.
+
+    This plots show the binned time/frequency interval between to spikes for neurons.
+
+    Args:
+        use_frequency(bool): use inverse interspike interval times (Hz)
+        binsize(float): bin size (milliseconds or Hz)
```
haleepfl, line 188: What does the default of `None` mean? Include `None` in the type.

It doesn't seem we enforce `binsize` as an `int` when using time? This is inconsistent with `time_binsize` above.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,363 @@
...
+    anim = FuncAnimation(fig, update_animation, frames=frames)
+    return anim, ax
+
+
+def frame_trace(filtered_report, plot_type='mean', ax=None):  # pragma: no cover
+    """Return potential plot displaying the voltage as a function of time from a frame report.
+
+    Args:
+        plot_type (str): string either `all` or `mean`. `all` will plot the first 15 traces from the
+        group. `mean` will plot the mean value of the node
```
haleepfl, line 329: Add an extra indent to this line so it is easier to distinguish the next arg from a line continuation.

bluepysnap/frame_report.py
```diff
@@ -111,6 +112,59 @@ def get(self, group=None, t_start=None, t_stop=None):
...
+        self.t_start = t_start
+        self.t_stop = t_stop
+
+    @cached_property
+    def report(self):
+        """Access to the report data.
+
+        Returns:
+            pandas.DataFrame: A DataFrame containing the data from the report. Row's index is the
+            different timestamps and the column's MultiIndex is :
```
haleepfl, line 143: Better to add an extra indentation on line continuations for args and return values

bluepysnap/spike_report.py
```diff
@@ -93,33 +94,79 @@ def get(self, group=None, t_start=None, t_stop=None):
...
+        self.t_start = t_start
+        self.t_stop = t_stop
+
+    @cached_property
+    def report(self):
+        """Access to the report data.
+
+        Returns:
+            pandas.DataFrame: A DataFrame containing the data from the report. Row's index is the
+            different timestamps and the columns are ids and population names.
```
haleepfl, line 147: Indent

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,363 @@
...
+
+    This animations represents the spiking nodes during time in a frame corresponding to the
+    x, y or z axis of the circuit.
+
+    Args:
+        x_axis (str): Node enum that will determine the animation x_axis
+        y_axis (str): Node enum that will determine the animation y_axis
+        dt (int) : the time bin size of each frame in the video in ms
+        ax(matplotlib.Axis): matplotlib Axis to draw on (if not specified, pyplot.gca()
+        and plt.figure() are used).
```
haleepfl, line 240: Indent

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,363 @@
...
+        dots[0].set_data(x, y)
+        return dots
+
+    frames = list(range(int(data.index[0] / dt), int(data.index[-1] / dt)))
+    anim = FuncAnimation(fig, update_animation, frames=frames)
+    return anim, ax
+
+
+def frame_trace(filtered_report, plot_type='mean', ax=None):  # pragma: no cover
+    """Return potential plot displaying the voltage as a function of time from a frame report.
```
lguyot, line 332: "potential plot" is a bit ambiguous: it may return something that is potentially a plot.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,363 @@
...
+        group. `mean` will plot the mean value of the node
+        ax: A plot axis object that will be updated
+
+    Returns:
+        matplotlib.Axis: axis containing the soma's traces.
+    """
+    # pylint: disable=too-many-locals
+
+    plt = _get_pyplot()
+    max_per_pop = 15
```
lguyot, line 345: Could be defined next to the place it is used for the first time.

bluepysnap/frame_report.py
```diff
@@ -111,6 +112,59 @@ def get(self, group=None, t_start=None, t_stop=None):
...
+    def __init__(self, frame_report, group=None, t_start=None, t_stop=None):
+        """Initialize a FilteredFrameReport.
+
+        A FilteredFrameReport is a lazy and cached object which contains the filtered data
+        from all the populations of a report.
+
+        Args:
+            frame_report (FrameReport): The FrameReport to filter.
+            group (None/int/list/np.array/dict): Get spikes filtered by group. See NodePopulation.
+            t_start (float): Include only frames occurring after this time.
```
lguyot, line 126: Strictly speaking, the times t_start and t_stop would be excluded. Is it the case?

bluepysnap/frame_report.py
lguyot: singular vs plural issue: "Row's index is the ... timestamps".

bluepysnap/frame_report.py
```diff
@@ -199,6 +253,22 @@ def __iter__(self):
...
 
+    def filter(self, group=None, t_start=None, t_stop=None):
+        """Returns a FilteredFrameReport.
+
+        A FilteredFrameReport is a lazy and cached object which contains the filtered data
+        from all the populations of a report.
+
+        Args:
+            group (None/int/list/np.array/dict): Get spikes filtered by group. See NodePopulation.
+            t_start (float): Include only frames occurring after this time.
```
lguyot, line 264: Same remark as before.

bluepysnap/frame_report.py
```diff
@@ -184,7 +238,7 @@ def simulation(self):
     @cached_property
     def population_names(self):
         """Returns the population names included in this report."""
-        return self._frame_reader.get_populations_names()
+        return sorted(self._frame_reader.get_populations_names())
```
lguyot, line 241: Should be reflected in a unit test (I'll check).

bluepysnap/_plotting.py
tomdele: That's fair. I will use "node_id" (without "s") which is the name in specs.

bluepysnap/_plotting.py
tomdele: Totally acceptable. 
I changed it for None/int/float and added a comment about the None. 
I added a raise if time_binsize <= 0.

bluepysnap/_plotting.py
tomdele: Indeed. I changed it. 
It is now consistent with the one above.

bluepysnap/_plotting.py
tomdele: done

bluepysnap/frame_report.py
tomdele: ok.

bluepysnap/spike_report.py
tomdele: done

bluepysnap/_plotting.py
tomdele: done

bluepysnap/nodes.py
```diff
@@ -239,7 +248,8 @@ def _positional_mask(self, node_ids):
         if node_ids is None:
             return np.full(len(self._data), fill_value=True)
         mask = np.full(len(self._data), fill_value=False)
-        mask[node_ids] = True
+        valid_node_ids = pd.Index(utils.ensure_list(node_ids)).intersection(self._data.index)
```
lguyot, line 251: Should be reflected in unit test (I'll check).

bluepysnap/spike_report.py
```diff
@@ -93,33 +94,79 @@ def get(self, group=None, t_start=None, t_stop=None):
         """Fetch spikes from the report.
 
         Args:
-            group (int/list/np.array/dict): Get spikes filtered by group. See NodePopulation.
+            group (None/int/list/np.array/dict): Get spikes filtered by group. See NodePopulation.
             t_start (float): Include only spikes occurring after this time.
```
lguyot, line 98: Same comment as before about inclusion/exclusion of bounds.

bluepysnap/spike_report.py
```diff
@@ -93,33 +94,79 @@ def get(self, group=None, t_start=None, t_stop=None):
...
+    def __init__(self, spike_report, group=None, t_start=None, t_stop=None):
+        """Initialize a FilteredSpikeReport.
+
+        A FilteredSpikeReport is a lazy and cached object which contains the filtered data
+        from all the populations of a report.
+
+        Args:
+            spike_report (SpikeReport): The SpikeReport to filter.
+            group (None/int/list/np.array/dict): Get spikes filtered by group. See NodePopulation.
+            t_start (float): Include only frames occurring after this time.
```
lguyot, line 130: idem

bluepysnap/spike_report.py
```diff
@@ -93,33 +94,79 @@ def get(self, group=None, t_start=None, t_stop=None):
...
+        self.group = group
+        self.t_start = t_start
+        self.t_stop = t_stop
+
+    @cached_property
+    def report(self):
+        """Access to the report data.
+
+        Returns:
+            pandas.DataFrame: A DataFrame containing the data from the report. Row's index is the
```
lguyot, line 146: plural versus singular issue: "Row's index is ... timestamps".

bluepysnap/spike_report.py
```diff
@@ -180,7 +227,7 @@ def _spike_reader(self):
     @cached_property
     def population_names(self):
         """Returns the population names included in this report."""
-        return self._spike_reader.get_populations_names()
+        return sorted(self._spike_reader.get_populations_names())
```
lguyot, line 230: Should be reflected in unit tests (I'll check).

bluepysnap/spike_report.py
```diff
@@ -194,3 +241,19 @@ def __getitem__(self, population_name):
...
+
+    def filter(self, group=None, t_start=None, t_stop=None):
+        """Returns a FilteredSpikeReport.
+
+        A FilteredSpikeReport is a lazy and cached object which contains the filtered data
+        from all the populations of a report.
+
+        Args:
+            group (None/int/list/np.array/dict): Get spikes filtered by group. See NodePopulation.
+            t_start (float): Include only frames occurring after this time.
```
lguyot, line 253: Same remarks about inclusion/exclusion of bounds.

bluepysnap/_plotting.py
haleepfl: Suggestion: `Return a plot displaying the voltage...` since "potential" is redundant here.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,363 @@
...
+    ax.hist(values, bins=bins, edgecolor='black', density=True)
+    return ax
+
+
+def spikes_firing_animation(filtered_report, x_axis=Node.X, y_axis=Node.Y,
+                            dt=20, ax=None):  # pragma: no cover
+    # pylint: disable=too-many-locals,too-many-arguments,anomalous-backslash-in-string
+    """Simple animation of simulation spikes.
+
+    This animations represents the spiking nodes during time in a frame corresponding to the
```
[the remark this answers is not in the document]
tomdele: Changed it for:
Each frame of the animation represents the spiking nodes during a period of dt ms seconds
    in a coordinate system corresponding to the x, y or z axis of the circuit.

bluepysnap/_plotting.py
tomdele: changed

bluepysnap/_plotting.py
tomdele: done

bluepysnap/frame_report.py
tomdele: no indeed this is >=. 
But this is tricky since we talk about floats here. 
I changed the docstrings accordingly

bluepysnap/frame_report.py
tomdele: done

bluepysnap/frame_report.py
tomdele: this is implicitly done in the :
```
    def test_iter(self):
        assert sorted(list(self.test_obj)) == ["default", "default2"]
        for report in self.test_obj:
            isinstance(report, test_module.PopulationCompartmentReport)
```

bluepysnap/nodes.py
tomdele: this is done already:
```
npt.assert_equal(_call({"node_id": 1}), [1])
npt.assert_equal(_call({"node_id": [1]}), [1])
npt.assert_equal(_call({"node_id": [1, 2]}), [1, 2])
npt.assert_equal(_call({"node_id": [1, 2, 42]}), [1, 2])
```

bluepysnap/spike_report.py
tomdele: done

bluepysnap/spike_report.py
tomdele: done

bluepysnap/spike_report.py
tomdele: done here :
```
    def test_iter(self):
        assert sorted(list(self.test_obj)) == ["default", "default2"]
        for spikes in self.test_obj:
            isinstance(spikes, test_module.PopulationSpikeReport)
```

bluepysnap/spike_report.py
tomdele: done

bluepysnap/spike_report.py
tomdele: done

bluepysnap/frame_report.py
tomdele: changed

tomdele:
since the review is quite big, I would be please if @asanin-epfl or @mgeplf can also have a look.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,372 @@
...
+    return plt
+
+
+def spikes_firing_rate_histogram(filtered_report, time_binsize=None, ax=None):  # pragma: no cover
+    """Spike firing rate histogram.
+
+    This plot shows the number of nodes firing during a range of time.
+
+    Args:
+        time_binsize(None/int/float): bin size (milliseconds). If None, a binning heuristic is used
```
asanin-epfl, line 48: Describe `filtered_report` arg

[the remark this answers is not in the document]
tomdele: It is indeed. 
This whole module is a trick to be able to decorrelate as much as I can the plotting part from the logic/query part of snap (while being able to do some direct call to the plots from the query objects).
If you check inside the `spike_report.py` or `frame_report.py` I link these functions via: 
`raster = bluepysnap._plotting.spike_raster`
inside the scope of the class. 
Meaning, the first argument is actually equivalent to `self` and should not appear inside the docstring. The whole plotting module is private and should not be used directly so noone should notice the missing argument in docstrings here.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,372 @@
...
+    """
+    # pylint: disable=too-many-locals
+    plt = _get_pyplot()
+    if time_binsize <= 0:
+        raise BluepySnapError("Invalid time_binsize = {}. Should be > 0.".format(time_binsize))
+
+    spike_report = filtered_report.spike_report
+
+    times = filtered_report.report.index
+    node_count = filtered_report.report[['ids', 'population']].drop_duplicates().shape[0]
```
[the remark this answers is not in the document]
tomdele: from what I found this is the fastest way of counting the unique rows. 
Unique is used for series.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,372 @@
...
+    time_start = np.min(times)
+    time_stop = np.max(times)
+
+    if time_binsize is None:
+        # heuristic for a nice bin size (~100 spikes per bin on average)
+        time_binsize = min(50.0, (time_stop - time_start) / ((len(times) / 100.) + 1.))
+
+    bins = np.append(np.arange(time_start, time_stop, time_binsize), time_stop)
+    hist, bin_edges = np.histogram(times, bins=bins)
+    freq = 1.0 * hist / node_count / (0.001 * time_binsize)
```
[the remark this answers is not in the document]
tomdele: I need to verify more carefully. This was from bluepy.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,372 @@
...
+            data.loc[mask] = report.loc[mask, "ids"] + props["node_id_offset"]
+        else:
+            ids = report.loc[mask, "ids"].to_numpy()
+            try:
+                ys = spikes.nodes.get(properties=y_axis)
+            except BluepySnapError:
+                continue
+            # astype is used to avoid problems with the categorical
+            data[mask] = ys[ids].astype(dtype).to_numpy()
+        _update_raster_properties()
```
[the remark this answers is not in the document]
tomdele: It cracks me too...
what was your crack ? Maybe I can comment more so it becomes less a "write-only code".
The whole plot is a mess to create properly ...

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,372 @@
...
+    bins = np.append(np.arange(time_start, time_stop, time_binsize), time_stop)
+    hist, bin_edges = np.histogram(times, bins=bins)
+    freq = 1.0 * hist / node_count / (0.001 * time_binsize)
+
+    if ax is None:
+        ax = plt.gca()
+        ax.set_xlabel('Time [ms]')
+        ax.set_ylabel('PSTH [Hz]')
+
+    ax.plot(0.5 * (bin_edges[1:] + bin_edges[:-1]), freq, label="PSTH", drawstyle='steps-mid')
```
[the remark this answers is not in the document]
tomdele: This is to take the middle of the bin instead of the beginning of the bin. 
I added a comment.

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,372 @@
...
+        ax.set_title('time = ' + str(frame * dt) + ' ms')
+        mask = (data.index >= frame * dt) & (data.index <= (frame + 1) * dt)
+        positions = data.loc[mask, [x_axis, y_axis]].values
+        x = positions[:, 0]
+        y = positions[:, 1]
+        dots[0].set_data(x, y)
+        return dots
+
+    frames = list(range(int(data.index[0] / dt), int(data.index[-1] / dt)))
+    anim = FuncAnimation(fig, update_animation, frames=frames)
```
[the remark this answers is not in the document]
tomdele: Because matplotlib is utterly bad and works with some random global contexts.
You need to return the ax AND the anim if you want the animation to run. If you return the ax only you gonna have only the first frame. Don't ask why...

bluepysnap/_plotting.py
```diff
@@ -0,0 +1,372 @@
...
+import pandas as pd
+
+from bluepysnap.exceptions import BluepySnapError
+from bluepysnap.sonata_constants import Node
+from bluepysnap.utils import roundrobin
+
+L = logging.getLogger(__name__)
+
+
+def _get_pyplot():
```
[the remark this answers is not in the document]
tomdele: This is again to decorrelate as much as I can the plotting from the core while being able to call plotting function directly from the filtered objects. 
I want to keep the maplotlib as an extra in the setup.py BUT I need to import _plotting in the frame and spike report module. + I also want the plotting to be a minimal pollution in the report module.
So if you want to do that, you need to encapsulate the imports for matplotlib inside a function that is not run when imported in another module.

Then if you try to call a plot function without matplotib you will have the error message asking you to install it.

bluepysnap/frame_report.py
```diff
@@ -111,6 +112,59 @@ def get(self, group=None, t_start=None, t_stop=None):
...
+        for population in self.frame_report.population_names:
+            frames = self.frame_report[population]
+            try:
+                ids = frames.nodes.ids(group=self.group)
+            except BluepySnapError:
+                continue
+            data = frames.get(group=ids, t_start=self.t_start, t_stop=self.t_stop)
+            if data.empty:
+                continue
+            population = [population]
```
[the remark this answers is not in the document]
tomdele: sure done

bluepysnap/frame_report.py
```diff
@@ -111,6 +112,59 @@ def get(self, group=None, t_start=None, t_stop=None):
...
+            new_index = tuple(tuple(population + ensure_list(x)) for x in data.columns)
+            data.columns = pd.MultiIndex.from_tuples(new_index)
+            # need to do this in order to preserve MultiIndex for columns
+            res = data if res.empty else data.join(res, how='outer')
+        return res.sort_index().sort_index(axis=1)
+
+    # pylint: disable=protected-access
+    trace = bluepysnap._plotting.frame_trace
```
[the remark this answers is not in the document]
tomdele: indeed. The goal is to be able to: 
```
filtered_report = sim.reports["soma"].filter(group={"mtype": "SLM_PPA"})
filtered_report.trace(plot_type="all")
```
This is a pattern I adapted from the plotting in pandas

tomdele:
>There should be a monument for your work here.
I want it big and high ;)

bluepysnap/frame_report.py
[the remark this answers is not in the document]
asanin-epfl: good to know

tomdele:
>For me it was a very educational review.

That s a nice compliment. Thx.

tomdele:
Thank you all: @lguyot, @haleepfl, @jdcourcol, @mgeplf, @asanin-epfl for the review and/or the help.

PR #47, opened 2020-03-25
Node sets

tomdele:
Rework of the node sets. 

* Allows the `population` key
* Better handling of the `node_id` key
* Add the `compounds `node sets
* Add the `"$and"`, `"$or"` operators 

We had to create the `"$or"` to fulfil the combined node set requirements so we decided to provide it as a part of the possible queries.


bluepysnap/node_sets.py
```diff
@@ -0,0 +1,99 @@
...
+    if set_value is None:
+        raise BluepySnapError("Missing node_set: '{}'".format(set_name))
+    if not isinstance(set_value, (collections.Mapping, list)) or not set_value:
+        raise BluepySnapError("Ambiguous node_set: '{}'".format({set_name: set_value}))
+
+    # keep the content intact
+    set_value = deepcopy(set_value)
+
+    if isinstance(set_value, collections.Mapping):
+        return _sanitize(set_value)
```
wizmer: What about always adding to `resolved` what you are about to return ?
This way, you are sure that `resolved` is fully resolved when exiting the function. Otherwise, you still have to update it after it has run, as you do line 74 or line 65.

bluepysnap/nodes.py
```diff
@@ -239,9 +200,103 @@ def _check_ids(self, node_ids):
...
+        """Check if a property exists inside the dataset."""
         if prop not in self.property_names:
             raise BluepySnapError("No such property: '%s'" % prop)
 
+    def _get_node_set(self, node_set_name):
+        """Returns the node set named 'node_set_name'."""
+        if node_set_name not in self._node_sets:
+            raise BluepySnapError("Undefined node set: '%s'" % node_set_name)
+        res = self._node_sets[node_set_name]
+        return None if res == {} else res
```
wizmer: Why is empty set a special case ?
Maybe the `if res == {} then return None` should be in `def _node_sets` ?

bluepysnap/nodes.py
```diff
@@ -239,9 +200,103 @@ def _check_ids(self, node_ids):
...
+
+    def _positional_mask(self, node_ids):
+        """Positional mask for the node IDs.
+
```
wizmer: Add the special case of None in the docsting. I was expecting that it would return an all False mask.

bluepysnap/nodes.py
```diff
@@ -239,9 +200,103 @@ def _check_ids(self, node_ids):
...
+        if populations is not None and self.name not in set(utils.ensure_list(populations)):
+            node_ids = []
+        else:
+            node_ids = queries.pop(NODE_ID_KEY, None)
+        return queries, self._positional_mask(node_ids)
+
+    def _properties_mask(self, queries):
+        """Return mask of node IDs with rows matching `props` dict."""
+        # pylint: disable=assignment-from-no-return
+        unknown_props = set(queries) - set(set(self._data.columns) | {POPULATION_KEY, NODE_ID_KEY})
```
wizmer: Maybe this is clearer ? (maybe ?)
```suggestion
        unknown_props = set(queries) - set(self._data.columns) - {POPULATION_KEY, NODE_ID_KEY}
```

bluepysnap/node_sets.py
```diff
@@ -0,0 +1,99 @@
...
+# This library is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+"""Access to node set data."""
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/node_sets.py
```diff
@@ -0,0 +1,99 @@
...
+
+import collections
+from copy import deepcopy
+
+import numpy as np
+from bluepysnap.exceptions import BluepySnapError
+from bluepysnap import utils
+
+
+def _sanitize(node_set):
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/node_sets.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Notes:
+        if the node set is a combined node set then the function resolve all the node sets
+        included in the combined one.
+    """
+    if set_name in resolved:
+        # return already resolved node_sets
+        return resolved[set_name]
+
+    set_value = content.get(set_name)
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/node_sets.py
```diff
@@ -0,0 +1,99 @@
...
+    if set_value is None:
+        raise BluepySnapError("Missing node_set: '{}'".format(set_name))
+    if not isinstance(set_value, (collections.Mapping, list)) or not set_value:
```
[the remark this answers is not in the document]
tomdele: To prevent the empty node sets: `{}` and `[]`

bluepysnap/node_sets.py
```diff
@@ -0,0 +1,99 @@
...
+        raise BluepySnapError("Ambiguous node_set: '{}'".format({set_name: set_value}))
+
+    # keep the content intact
+    set_value = deepcopy(set_value)
```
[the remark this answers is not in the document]
tomdele: indeed. Done

bluepysnap/node_sets.py
tomdele: indeed I changed it

bluepysnap/nodes.py
tomdele: No this is from an older impl of the node sets where `{}` was interpreted as a None for ids.
I do not allow the `{}` anymore inside the node sets files

bluepysnap/nodes.py
tomdele: Done.
Yes the None behavior is due to the `group=None` behavior for `ids()`

bluepysnap/nodes.py
tomdele: yes makes sense. Done

bluepysnap/nodes.py
```diff
@@ -239,9 +200,105 @@ def _check_ids(self, node_ids):
...
+            raise BluepySnapError("Unknown node properties: [{0}]".format(", ".join(unknown_props)))
+
+        queries, mask = self._node_population_mask(queries)
+        if not mask.any():
+            # Avoid fail and/or processing time if wrong population or no nodes
+            return mask
+
+        for prop, values in six.iteritems(queries):
+            prop = self._data[prop]
+            if issubclass(prop.dtype.type, np.floating):
```
mgeplf, line 252: np.issubdtype ?

bluepysnap/nodes.py
```diff
@@ -239,9 +200,105 @@ def _check_ids(self, node_ids):
...
+        elif first_key == '$and':
+            queries = queries.pop("$and")
+            operator = np.logical_and
+        else:
+            return self._properties_mask(queries)
+
+        mask = np.full(len(self._data), first_key != "$or")
+        for query in queries:
+            mask = operator(mask, self._operator_mask(query))
+        return mask
```
wizmer, line 281: Optional and more of a fun fact, this is a reduce operation:
```python
return functools.reduce(lambda mask, query: operator(mask, self._operator_mask(query)),
                            queries, 
                            initializer=np.full(len(self._data), first_key != "$or"))
```
```

File: bluepysnap/nodes.py (hunk 10 above)
Comment (tomdele) [reply bluepysnap/nodes.py:252]: done

File: bluepysnap/nodes.py [hunk 12]
```diff
@@ -239,9 +200,105 @@ def _check_ids(self, node_ids):
...
+                prop_mask = _complex_query(prop, values)
+            else:
+                prop_mask = np.in1d(prop, values)
+            mask = np.logical_and(mask, prop_mask)
+        return mask
+
+    def _operator_mask(self, queries):
+        """Handle the query operators '$or', '$and'."""
+        # will pop the population and or/and operators so need to copy
+        queries = deepcopy(queries)
```
wizmer, line 267: You would not need the copy if you would use
`queries = queries[1:]` instead of the `pop`, no ?

bluepysnap/nodes.py
```diff
@@ -266,28 +323,33 @@ def ids(self, group=None, limit=None, sample=None):
...
+            >>> nodes.ids(group={ Node.LAYER: [2, 3]})  # returns list of IDs with layer in [2,3]
+            >>> nodes.ids(group={ Node.X: (0, 1)})  # returns list of IDs with 0 < x < 1
+            >>> # returns list of IDs matching one of the queries inside the 'or' list
+            >>> nodes.ids(group={'$or': [{ Node.LAYER: [2, 3]},
+            >>>                          { Node.X: (0, 1), Node.MTYPE: 'L1_SLAC' }]})
+            >>> # returns list of IDs matching all the queries inside the 'and' list
+            >>> nodes.ids(group={'$and': [{ Node.LAYER: [2, 3]},
+            >>>                           { Node.X: (0, 1), Node.MTYPE: 'L1_SLAC' }]})
         """
         # pylint: disable=too-many-branches
```
wizmer, line 344: You probably no longer need the pragma

bluepysnap/node_sets.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class NodeSets:
+    """Access to node sets data."""
+
+    def __init__(self, filepath):
+        """Initializes a node set object from a node sets file.
+
+        Args:
+            filepath (str): Path to a SONATA node sets file.
```
wizmer, line 111: It think pathlib.Path also works

tests/test_node_sets.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def test_get(self):
+        assert self.test_obj['Node2_L6_Y'] == {'mtype': 'L6_Y', 'node_id': [20, 30]}
+        assert self.test_obj['double_combined'] == {'$or': [{'$or': [
+                                              {'mtype': 'L6_Y', 'node_id': [20, 30]}, {'layer': [2, 3]}]},
+                                                                      {'population': 'default',
+                                                                       'mtype': 'L6_Y'}]}
+
+    def test_iter(self):
+        expected = sorted(list(json.load(open(str(TEST_DATA_DIR / 'node_sets_file.json')))))
```
wizmer, line 50: sorted(list -> set 
no ?

bluepysnap/nodes.py
tomdele: then it becomes a destructive operation for queries which I don t want.

bluepysnap/node_sets.py
tomdele: indeed

bluepysnap/node_sets.py
```diff
@@ -0,0 +1,125 @@
...
+        raise BluepySnapError("Ambiguous node_set: '{}'".format({node_set_name: set_value}))
+    if isinstance(set_value, collections.Mapping):
+        resolved[node_set_name] = _sanitize(set_value)
+        return resolved[node_set_name]
+
+    # compounds only
+    res = []
+    for sub_set_name in set_value:
+        sub_res_dict = _resolve_set(content, resolved, sub_set_name)
+        res.append(sub_res_dict)
```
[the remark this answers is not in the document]
tomdele: yep missed this one

bluepysnap/nodes.py
tomdele: nice one but I think it is a bit easier to read with the for loop ....

bluepysnap/nodes.py
tomdele: indeed

tomdele:
Thank you all for the very nice review. Really appreciated.

PR #51, opened 2020-04-08
Fix empty list in reports

tomdele:
* this is a temporary solution until
  https://github.com/BlueBrain/libsonata/issues/84 is done

### DIFF
--- a/bluepysnap/frame_report.py
+++ b/bluepysnap/frame_report.py
@@ -19,6 +19,7 @@ import logging
 
 from cached_property import cached_property
 from pathlib2 import Path
+import numpy as np
 import pandas as pd
 from libsonata import ElementReportReader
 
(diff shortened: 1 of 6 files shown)


bluepysnap/frame_report.py
```diff
@@ -211,6 +212,10 @@ def nodes(self):
 
     def _resolve(self, group):
         """Transform a group into a node_id array."""
+        if group == []:
+            # Only solution to return an empty list from self._frame_population.get
+            # see: https://github.com/BlueBrain/libsonata/issues/84
+            return np.array([-2])
```
wizmer, line 218: You could define

```python
def empty_list():
    '''Only solution to return an empty list from self._frame_population.get
       see: https://github.com/BlueBrain/libsonata/issues/84
    '''
   return np.array([-2])
```
And use it everywhere, so that you won't have to look for np.array([-2]) the day you need to fix it, and you don't need to rewrite the comment everytime you need it.

tomdele: yes why not.

PR #52, opened 2020-04-08
fix report query for empty arrays or tuple

tomdele:
fix for report .get function with empty tuple and array

### DIFF
--- a/bluepysnap/frame_report.py
+++ b/bluepysnap/frame_report.py
@@ -19,6 +19,7 @@ import logging
 
 from cached_property import cached_property
 from pathlib2 import Path
+import numpy as np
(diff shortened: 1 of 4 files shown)


asanin-epfl:
Should we check just for an instance of iterable then?

tomdele:
No, because dicts are also iterables and the behavior of `nodes.ids(group={})` is: return all node_ids. So I wanted to keep the same behavior here. 
I never found a proper function included in a widespread lib to check this kind of stuffs... So if you have something to do that in a clean way I take it !

tomdele:
simple question but not so simple answer :smile:

tomdele:
thank you guys

PR #56, opened 2020-04-16
Use categoritical values for attr in @library

tomdele:
* use categorical for attr in library
* add a test file containing library


tomdele:
see: 
https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#nodes---enum-datatypes

PR #57, opened 2020-04-29
Fixing the iter_connection with unique_node_ids

tomdele:
* fixing the iter_connection function
* adding comments and explanations in docstrings
* fixing the pylint in tox due to the multiprocessing error
  https://github.com/PyCQA/pylint/issues/3524


tests/test_edges.py
```diff
@@ -420,3 +420,27 @@ def test_iter_connections_10(self):
             self.test_obj.iter_connections(
                 [0, 2], [1], return_edge_ids=True, return_edge_count=True
             )
+
+    def test_iter_connection_unique(self):
+        test_obj = TestEdgePopulation.create_population(
+            str(TEST_DATA_DIR / "edges_complete_graph.h5"), 'default')
+        it = test_obj.iter_connections([0, 1, 2], [0, 1, 2])
+        assert sorted(it) == [(0, 1), (0, 2), (1, 0), (1, 2), (2, 0), (2, 1)]
```
asanin-epfl, line 428: This is tricky! What would happen if there is `(0, 0)` in the original result of `iter_connections(unique_node_ids=False)`?

tomdele: This will still exist.
We check both ways separately source and target.

tomdele:
whoo I missed your comment from weeks ago ... sorry about that @asanin-epfl

tests/test_edges.py
tomdele: This first test is just to verify that the input data are still what we expect.

tomdele: with a 0-selfloop the results would be something like this:
```
test_obj.iter_connections([0, 1, 2], [0, 1, 2], unique_node_ids=True)
> [(0, 0),  (1, 2), (2, 1)]
```

asanin-epfl: But in your tests you have different results:
```
        it = test_obj.iter_connections([0, 1, 2], [0, 1, 2], unique_node_ids=True)
        assert sorted(it) == [(0, 1), (1, 0)]
```

tomdele: yes because I do not have a self loop with 0.

tomdele:
```
it = test_obj.iter_connections([0, 1, 2], [0, 1, 2], unique_node_ids=True)
assert sorted(it) == [(0, 1), (1, 0)]
```
in this case: 
0 is not used as a source so I can use it 
1 is not used as a target so I can use it 
(0, 1)
then I move to 1 as source:
1 is not used as a source so I can use it 
0 is not used a target so I can use it
(0, 1)
then I move to 2 as source:
2 is not used as a source so I can use it
0 is used a target so I cannot use it
1 is used a target so I cannot use it
there is no (2, x)
results = 
[(0, 1), (1, 0)]

PR #60, opened 2020-05-27
Fix raster and trace

tomdele:
Quick fix for the trace and raster plots


tomdele:
Yes 100%.

tomdele:
Except if the simulators change massively but then I cannot predict anything ...

tomdele:
Thank you !

PR #63, opened 2020-06-17
Fix documentation build with new Sphinx

haleepfl:
New sphinx appears to have an issue with the "name" template
variable. To solve, we remove our custom template which will
also reduce the risk of future breakages. This will however
add the fully qualified submodule name in the toctree.


tomdele:
I must say I am not sure what I review here ...

tomdele:
Can you give me more info ?

haleepfl:
SNAP documentation builds are currently failing: https://readthedocs.org/projects/bluebrainsnap/builds/11240661/

This seems to be caused by a new release of Sphinx being incompatible with our custom autosummary template. To simplify things, I've remove the custom template (at the expense of slightly longer tables of contents) which fixes the build issue.

tomdele:
Ok, thanks for the clarification. 
So we gonna have the attributes and the raise also ?

haleepfl:
> Ok, thanks for the clarification.
> So we gonna have the attributes and the raise also ?

Yes, we should see all module members.

PR #64, opened 2020-06-18
Fix circuit validation for missing edge_group_id/index

tomdele:
Both fields are not mandatory and actually not supported by
libsonata and snap. These fields are not mandatory and using them would
result in a huge perf loss.

* circuits are valid if both edge_group_id/index are missing and group
  count is equal to 1.

### DIFF
--- a/bluepysnap/circuit_validation.py
+++ b/bluepysnap/circuit_validation.py
@@ -169,14 +169,14 @@ def _find_nodes_population(node_population_name, nodes):
     return None
 
 
-def _get_group_name(group, parents=1):
+def _get_group_name(group, parents=0):
     """Gets group name of h5 group.
 
     Args:
         group (h5py.Group): nodes group in nodes .h5 file
         parents (int): number of extra parents needed
     """
-    return Path(*Path(group.name).parts[-(parents+1):])
+    return Path(*Path(group.name).parts[-(parents + 1):])
 
 
 def _get_h5_data(h5, path):
@@ -224,7 +224,7 @@ def _check_bio_nodes_group(group, config):
                                        format(group_name, group.file.filename)))
 
     errors = []
-    group_name = _get_group_name(group)
+    group_name = _get_group_name(group, parents=1)
     missing_fields = sorted({'morphology', 'x', 'y', 'z'} - set(group))
     if missing_fields:
         errors.append(fatal('Group {} of {} misses biophysical fields: {}'.
@@ -262,7 +262,8 @@ def _check_nodes_group(group, config):
     missing_fields = sorted(set(REQUIRED_GROUP_NAMES) - set(group))
     if missing_fields:
         return [fatal('Group {} of {} misses required fields: {}'
-                      .format(_get_group_name(group), group.file.filename, missing_fields))]
+                      .format(_get_group_name(group, parents=1), group.file.filename,
+                              missing_fields))]
     elif 'biophysical' in group['model_type'][:]:
         return _check_bio_nodes_group(group, config)
     return []
@@ -320,7 +321,8 @@ def _check_edges_group_bbp(group):
     missing_fields = sorted(set(GROUP_NAMES) - set(group))
     if missing_fields:
         return [BbpError(Error.WARNING, 'Group {} of {} misses fields: {}'.
-                         format(_get_group_name(group), group.file.filename, missing_fields))]
+                         format(_get_group_name(group, parents=1), group.file.filename,
+                                missing_fields))]
     return []
 
 
@@ -422,19 +424,19 @@ def _get_edge_population_groups(population):
 def _check_edge_group_id(population):
     errors = []
     groups = np.array(_get_edge_population_groups(population))
-    group_datasets = ["edge_group_id",  "edge_group_index"]
+    group_datasets = ["edge_group_id", "edge_group_index"]
     missing_datasets = set(group_datasets) - set(population)
-    population_name = _get_group_name(population, parents=0)
+    population_name = _get_group_name(population)
     if len(groups) > 1:
         errors.append(BbpError(Error.WARNING, 'Population {} of {} have multiple groups. '
                                               'Cannot be read via bluepysnap or libsonata'.
-                               format(population_name, population.file.filename, missing_datasets)))
+                               format(population_name, population.file.filename)))
     if len(missing_datasets) == 2 and len(groups) == 1:
         # no "edge_group_id", "edge_group_index" but only one group
         return errors
     elif len(missing_datasets) == 1:
         return errors + [fatal('Population {} of {} misses dataset {}'.
-                                   format(population_name, population.file.filename, missing_datasets))]
+                               format(population_name, population.file.filename, missing_datasets))]
 
     edge_group_ids = population["edge_group_id"][:]
     edge_group_index = population["edge_group_index"][:]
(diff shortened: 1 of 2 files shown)


tomdele:
Need to do the cleaning and linting

bluepysnap/circuit_validation.py
```diff
@@ -418,33 +472,43 @@ def _check_edges_population(edges_dict, nodes):
...
             return errors
+
         for population_name in edges:
+            required_datasets = ['edge_type_id', 'source_node_id', 'target_node_id']
+
             population_path = '/edges/' + population_name
             population = h5f[population_path]
+            groups = _get_edge_population_groups(population)
+
+            if len(groups) != 1:
```
asanin-epfl, line 490: In case there are no groups and `len`==0? Don't we already have this check inside of `_check_edge_group_id`?

bluepysnap/circuit_validation.py
```diff
@@ -408,6 +415,53 @@ def _check(indices, nodes_ds):
...
+                                              'Cannot be read via bluepysnap or libsonata'.
+                               format(population_name, population.file.filename)))
+    if len(missing_datasets) == 2 and len(groups) == 1:
+        # no "edge_group_id", "edge_group_index" and only one group --> can use implicit ids
+        return errors
+    elif len(missing_datasets) == 1:
+        return errors + [fatal('Population {} of {} misses dataset {}'.
+                               format(population_name, population.file.filename, missing_datasets))]
+
+    edge_group_ids = population["edge_group_id"][:]
```
asanin-epfl, line 441: Will we get here in case of `len(missing_datasets) == 2 and len(groups) > 1` ?

tests/test_circuit_validation.py
```diff
@@ -288,26 +290,146 @@ def test_no_edges_h5():
         assert errors == [Error(Error.FATAL, 'No "edges" in {}.'.format(edges_file))]
 
 
-def test_no_required_edge_population_datasets():
+def test_no_required_edge_population_datasets_one_group():
+    required_datasets = sorted([
+        'edge_type_id', 'source_node_id', 'target_node_id'])
+    with _copy_circuit() as (circuit_copy_path, config_copy_path):
```
asanin-epfl, line 296: should we also have here `optional_datasets = ['edge_group_id', 'edge_group_index']`? We will also delete them
```
            for ds in optional_datasets:
                del h5f['edges/default/' + ds]
```
And they should not appear as errors. Or should they?

tests/test_circuit_validation.py
```diff
@@ -288,26 +290,146 @@ def test_no_edges_h5():
...
+
+def test_edge_population_edge_group_different_length():
+    with _copy_circuit() as (circuit_copy_path, config_copy_path):
+        edges_file = circuit_copy_path / 'edges.h5'
+        with h5py.File(edges_file, 'r+') as h5f:
+            del h5f['edges/default/edge_group_index']
+            h5f.create_dataset('edges/default/edge_group_index', data=[0, 1, 2, 3, 4])
+        errors = test_module.validate(str(config_copy_path))
+        assert errors == [Error(Error.FATAL,
+                                'Population default of {} "edge_group_ids" and "edge_group_index" of different sizes'.
```
asanin-epfl, line 369: Should be "edge_group_id" instead of "edge_group_id**s**" ?

tests/test_circuit_validation.py
```diff
@@ -288,26 +290,146 @@ def test_no_edges_h5():
...
+
+
+def test_edge_population_wrong_group_id():
+    with _copy_circuit() as (circuit_copy_path, config_copy_path):
+        edges_file = circuit_copy_path / 'edges.h5'
+        with h5py.File(edges_file, 'r+') as h5f:
+            del h5f['edges/default/edge_group_id']
+            h5f.create_dataset('edges/default/edge_group_id', data=[0, 1, 0, 0])
+        errors = test_module.validate(str(config_copy_path))
+        print('Population default of {} misses group(s): {}'.
```
asanin-epfl, line 380: Remaining `print`?

asanin-epfl:
Can you please write down here a draft of logic of `edge_group_id`, `edge_group_index` validation? Should we also consider here `edge_id` field?

tomdele:
>Can you please write down here a draft of logic of edge_group_id, edge_group_index validation? Should we also consider here edge_id field?

Sure: 

- If there is only one group for this population, the `edge_group_id`, `edge_group_index` are not mandatory . Both can be inferred :  `edge_group_id` will be an array of length `N_Edge` with the name of the single group as values and the `edge_group_index` will be the positional index for datasets. 
- If there is more than 1 group, you must have both `edge_group_id` and `edge_group_index`
- If there is one dataset : either `edge_group_id` or `edge_group_index` this is not correct. You must have both (or none if you have one group)
- If you have both datasets : both should have the same length
- All groups in `edge_group_id` must exist for this population. 
- At least all values from the `edge_group_index` dataset can be accessed inside the group datasets. That is, the maximal value for a given group id in `edge_group_index` must be <= length of the property dataset inside the group. 


I did not test for the `edge_id` indeed maybe I should do it. But this is a pain and can take forever considering the size of the edge arrays ...

asanin-epfl:
Thank you! No need to test for `edge_id`. That would be too much. I just wanted to clarify.

tests/test_circuit_validation.py
tomdele: They should not appears as error. I think a test is missing.

bluepysnap/circuit_validation.py
tomdele: no you should raise at line 435

bluepysnap/circuit_validation.py
tomdele: Actually not. I need to take the "no group" case into account. 
This can maybe happens and should be a valid use case (you just want want the graph topology).

tests/test_circuit_validation.py
tomdele: indeed !

tests/test_circuit_validation.py
tomdele: yep !

bluepysnap/circuit_validation.py
asanin-epfl: Will we get into it if `len(groups) > 1`. The condition is for `len(groups)==1`.

bluepysnap/circuit_validation.py
```diff
@@ -418,33 +474,43 @@ def _check_edges_population(edges_dict, nodes):
...
             return errors
+
         for population_name in edges:
+            required_datasets = ['edge_type_id', 'source_node_id', 'target_node_id']
+
             population_path = '/edges/' + population_name
             population = h5f[population_path]
+            groups = _get_edge_population_groups(population)
+
+            if len(groups) > 1:
```
asanin-epfl, line 492: Would it be possible move this check inside of `check_edge_group_id` so all checks of `edge_group_id`/`edge_group_index` are in one place?

tomdele: sure doing it.

tomdele:
The code is dirty ... but that s a mess to separate the `_check_edge_population_data` into smaller functions (due to the mandatory returns in case of a missing dataset/group)

bluepysnap/circuit_validation.py
tomdele: we want to go there if (group >= 1 and len(missing_group_datasets)==0)

tomdele: if  (len(groups)>1 and len(missing_group_datasets)>0) I raise at `if missing_datasets:`  line 440

asanin-epfl:
`# pylint: disable=too-many-locals,too-many-return-statements,too-many-branches` that's a combo! :D

tomdele:
># pylint: disable=too-many-locals,too-many-return-statements,too-many-branches that's a combo! :D

........ I know :disappointed: 
I wanted to get rid of that but that was such a mess with the returns ... I accepted to add this magical line ...

tomdele:
I am a bit ashamed

asanin-epfl:
No reason to be ashamed. I don't know a better way. All good from me.

tomdele:
Thx for the review !

PR #67, opened 2020-06-23
fix for the biophysical node validation

tomdele:
The validation for the biophysical was missing for "in @library" use case. I added it.

### DIFF
--- a/bluepysnap/circuit_validation.py
+++ b/bluepysnap/circuit_validation.py
@@ -5,6 +5,7 @@ import numpy as np
 import click
 from pathlib2 import Path
 import h5py
+import six
 
 from bluepysnap.config import Config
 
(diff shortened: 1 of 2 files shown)


bluepysnap/circuit_validation.py
```diff
@@ -248,6 +249,18 @@ def _check_rotations():
     return errors
 
 
+def _is_biophysical(group):
+    """Check if a group contains biophysical nodes."""
+    if group['model_type'][0] == 'biophysical':
```
asanin-epfl, line 254: Can we at least put some warning here that we rely on homogeneity of group? If two different types are mixed in the same group => error here. Specs don't prohibit mixing.

tomdele: Actually I thought about this and I agree this is not forbidden by the specs but I concluded it does not make sense or it is impossible to have different model_type in the same group because they cannot share the same fields (which is the point of a group to me). 
- biophysical --> morphology, positions, rotations etc 
- virtual --> nothing
- single_compartment --> diameter, maybe positions
- point_neuron --> positions

what do you think ?

tomdele: I am not against putting a warning but if it is shown for every circuit, I think we should not put it

asanin-epfl: The solution according to the specs would be to use those node_type.csv files. Now we have to abuse so much storage for nothing.

tomdele: I really need to dig more into this csv files ... Unfortunately, I think you know more than I do about this ... I need to catch up.

asanin-epfl: I would suggest a separate PR for that. Here is good to go except this failed CI.

tomdele: Space-wise it is a bit better now we include this to the library.

PR #69, opened 2020-07-02
Add source/target to the available properties for edges

tomdele:


asanin-epfl:
looks good, some CI error for python2.7 on type comparison.

tomdele:
Yeah I have seen that. I couldn't find what s happening yet ... 
I will correct it.

PR #71, opened 2020-07-07
Add as_edges_source/as_edges_target to NodePopulations

tomdele:


tomdele:
I am not so sure about the naming and description... 
@asanin-epfl  do you think this is clear enough ?

bluepysnap/nodes.py
```diff
@@ -155,6 +155,16 @@ def _property_names(self):
     def _dynamics_params_names(self):
         return set(utils.add_dynamic_prefix(self._population.dynamics_attribute_names))
 
+    def as_edges_source(self):
+        """Set of edges population that use this node population as source."""
```
asanin-epfl: Missing plural **s** """Set of edges population**s**...

tomdele: thanks ! I changed it in the last commit.

tomdele:
Yes you are perfectly right about the naming. 
I feel like `source_edges` is not super clear. What about `source_in_edges` ?

asanin-epfl:
> What about `source_in_edges`?

Works good too. The prefix `as` only worried me.

tomdele:
perfect thanks for the feedback I will change this

PR #73, opened 2020-07-08
Reducing memory usage for fields from @library

tomdele:


tomdele:
This is something bench by @mgeplf  long time ago for bluepy

tomdele:
> very well done.

All credit goes to Arseny and Mike ;)

PR #74, opened 2020-07-08
Add a test to validate a missing morphology file in @library/morphology

asanin-epfl:


tomdele:
Don t you need to do this for `model_template` also ?

PR #76, opened 2020-07-13
Add the $target query to nodes

tomdele:
It can be useful to query node set using combined queries.
To do this a new field must be added to the queries.


tomdele:
used by https://github.com/BlueBrain/mooc-grading-service/issues/131

tomdele:
Thanks you !

PR #78, opened 2020-07-14
Don't raise errors when `node_id` is defined implicitly.

asanin-epfl:


tomdele:
Totally forgot about that ! 
That s cool !

asanin-epfl:
Me too. Thank you!

PR #79, opened 2020-07-16
Fix circuit validation of 'node_group_id' and 'node_group_index'

asanin-epfl:
'node_group_id' and 'node_group_index' must be required when multiple groups. Currently we require them when multiple populations.

### DIFF
--- a/bluepysnap/circuit_validation.py
+++ b/bluepysnap/circuit_validation.py
@@ -7,6 +7,7 @@ from pathlib2 import Path
 import h5py
 import six
 
+from bluepysnap import BluepySnapError
 from bluepysnap.config import Config
 
 MAX_MISSING_FILES_DISPLAY = 10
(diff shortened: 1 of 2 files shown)


bluepysnap/circuit_validation.py
```diff
@@ -303,17 +303,18 @@ def _check_nodes_population(nodes_dict, config):
...
             errors.append(fatal('No "nodes" in {}.'.format(nodes_file)))
             return errors
-        if len(nodes.keys()) > 1:
-            required_datasets += ['node_group_id', 'node_group_index']
         for population_name in nodes:
             population = nodes[population_name]
+            groups = [population[name] for name in population
+                      if isinstance(population[name], h5py.Group)]
+            if len(groups) > 1:
+                required_datasets += ['node_group_id', 'node_group_index']
```
tomdele, line 311: I do more checks for the edges iirc. 
Here https://github.com/BlueBrain/snap/pull/79/files#diff-ab66ad956573cb0b56e8d613adbc9ec6R448-R475
maybe this is worth it to implement ? Or do you cover all the possible usecases allready ?

asanin-epfl: Done, please verify

asanin-epfl:
Thank you! I've been exhausted by this PR.

tomdele:
Yeah, I can understand. I had the same problem when I added it for the edges ... 
Having all these possibilities we need to take into account and all the indirections is just a pain and takes a lot of attention.

PR #81, opened 2020-07-20
Adapt validation to sonata examples

asanin-epfl:


bluepysnap/circuit_validation.py
```diff
@@ -206,6 +207,40 @@ def _get_population_groups(population_h5):
             if isinstance(population_h5[name], h5py.Group) and name.isdigit()]
 
 
+def _sonata_group_to_dataframe(group, types_file, type_id, id_, group_id, group_index):
+    """Transforms hdf5 population group to pandas DataFrame.
+
+    Args:
+        group: HDF5 group of edges or nodes population
```
tomdele, line 214: I am pretty sure we cannot do that for edges. 
With a "normal" edge file (few hundred of millions of edges to multiple billions of edges), the data cannot fit in memory.
Did you try with a big circuit ? 

Maybe we should keep that for nodes only at the moment ?

bluepysnap/circuit_validation.py
```diff
@@ -206,6 +207,40 @@ def _get_population_groups(population_h5):
...
+            df[k] = v[:]
+    if '@library' in group:
+        for k, v in group['@library'].items():
+            if isinstance(v, h5py.Dataset):
+                df[k] = v[:][df[k].to_numpy(dtype=int)]
+    if types_file is None:
+        return df
+    types = pd.read_csv(types_file, sep=r'\s+')
+    types.rename(columns={types.columns[0]: 'type_id'}, inplace=True)
+    return pd.merge(df, types, on='type_id', how='left')
```
tomdele, line 241: that s cool it makes a first impl of the csv !

bluepysnap/circuit_validation.py
tomdele: Ah ok I saw that you are using it only for the nodes. 
Is it something you plan on using on edges too ?

asanin-epfl: I thought to use it for edges but now I'm thinking. It will definitely fail on few hundred of millions of edges. I will reimplement it only for nodes.

bluepysnap/circuit_validation.py
asanin-epfl: Now we can validate Allen circuits as they heavily rely on this `.csv`.

tomdele:
That s really cool. Thanks for making this.

Do we validate the AIBS circuit with these changes ?

asanin-epfl:
Yes. I did. Here is the log:
```
FATAL: Group external/0 of /Users/sanin/workspace/sonata/examples/300_cells/network/external_nodes.h5 misses required fields: ['model_template']
```

tomdele:
So it spots errors that s great !

asanin-epfl:
Thank for the review!

PR #83, opened 2020-07-27
Fix the sample with empty group NodePopulation.ids

tomdele:
* Fix the sample raise if the group return empty list and sample is used

### DIFF
--- a/bluepysnap/nodes.py
+++ b/bluepysnap/nodes.py
@@ -428,9 +428,9 @@ class NodePopulation(object):
             result = utils.ensure_list(group)
             self._check_ids(result)
             preserve_order = isinstance(group, collections.Sequence)
-
         if sample is not None:
-            result = np.random.choice(result, sample, replace=False)
+            if len(result) > 0:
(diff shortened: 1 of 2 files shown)


bluepysnap/nodes.py
```diff
@@ -428,9 +428,9 @@ def ids(self, group=None, limit=None, sample=None):
             result = utils.ensure_list(group)
             self._check_ids(result)
             preserve_order = isinstance(group, collections.Sequence)
-
         if sample is not None:
-            result = np.random.choice(result, sample, replace=False)
+            if len(result) > 0:
+                result = np.random.choice(result, sample, replace=False)
             preserve_order = False
```
asanin-epfl, line 434: Would it be correct to set `preserve_order` if `len(result) == 0?`

tomdele: The preserve_order change is really minimal because it will only result in calling the unique on an empty array and that s it. 
The factorization worth the very small overhead imao.

asanin-epfl: Thank you for clarifying.

PR #85, opened 2020-07-30
Change the config handling

tomdele:
Better paths handling in config
* Use pathlib instead of os.path
* manifest not mandatory anymore
* Better handling of the relative paths in manifest and in the different paths
   - can use all the "." + something (i.e: ./dir, ../, ./../, etc)
* strings resolved as path only if contains '$' or starts with '.'
* Add raises to avoid errors :
   - double abs path resulting in single abs path
   - manifest anchor with more than one sub-anchor
   - check if all the paths returned are absolute paths (when possible)
   - if an anchor (which is an abspath) is badly placed

### DIFF
--- a/bluepysnap/config.py
+++ b/bluepysnap/config.py
@@ -91,7 +91,7 @@ class Config(object):
             abs_paths = [v for v in vs[1:] if v.startswith('/')]
             if len(abs_paths) != 0:
                 raise BluepySnapError("Misplaced anchors in : {}."
-                                      "Please verify your '$' usage.".format(value, vs))
+                                      "Please verify your '$' usage.".format(value))
             return str(Path(*vs))
         elif value.startswith('.'):
             return str(Path(self.manifest['${configdir}'], value).resolve())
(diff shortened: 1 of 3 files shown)


asanin-epfl:
@jonathanlurie I will go after you here.

tomdele:
Rah need to fix the python2 first

bluepysnap/config.py
```diff
@@ -56,33 +57,46 @@ def _resolve_manifest(manifest, configdir):
...
-                result[k] = os.path.join(configdir, v[2:])
+            if v.startswith('.'):
+                result[k] = str(Path(configdir, v).resolve())
 
         while True:
             update = False
             for k, v in six.iteritems(result):
+                if v.count('$') > 1:
+                    raise BluepySnapError(
+                        '{} is not a valid anchor : contains more than one sub anchor.')
```
joni-herttuainen, line 68: missing `.format()`

tomdele: indeed !

tomdele:
Thank you guys !

PR #88, opened 2020-08-06
Update of the example notebooks

tomdele:
* changed the circuit from gpfs
* changed the notebooks accordingly


tomdele:
Yo @antonelepfl , 
Thanks to your issue I changed the notebooks and the circuit. 
Let me know if everything is good for you now !

antonelepfl:
I'm currently working on a transformation from BluePy to Snap of some of the analysis that I have to do so I'll be using snap and the plots. I'll let you know if I find any other issue soon
We can keep this PR open for a couple of days and then if there is not any other issue we close it.

PR #90, opened 2020-08-25
Fix deprecation warnings

### LINKED ISSUE #89: Deprecation warning collections
Just letting you know that my tests get a deprecation warning:
bluepysnap/nodes.py:421: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working

tirkarthi:
* Import ABC from collections.abc in Python 3. Fixes #89 
* Fix deprecation warning regarding invalid escape sequences.

### DIFF
--- a/bluepysnap/_plotting.py
+++ b/bluepysnap/_plotting.py
@@ -307,8 +307,8 @@ def spikes_firing_animation(filtered_report, x_axis=Node.X, y_axis=Node.Y,
         y_limits = [data[y_axis].min(), data[y_axis].max()]
         ax.set_xlim(*x_limits)
         ax.set_ylim(*y_limits)
-        ax.set_xlabel('{} $\mu$m'.format(x_axis))  # noqa
-        ax.set_ylabel('{} $\mu$m'.format(y_axis))  # noqa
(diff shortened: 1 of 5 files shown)


tirkarthi:
The CI seems to fail due to test coverage being less than 100% . I guess the branch of Python 2 code won't be run which might be affecting this.

mgeplf:
We will look at this next week, thanks for the contribution.

tomdele:
Hello, 
thanks @tirkarthi for the contribution. This is really appreciated ! 
Indeed the py27 try/except branch cannot be tested during the py36 testing run ... I never found a proper way of testing this for widely spread libraries during the tests.
Ex here, we cannot do a simple: 
```
def test_import():
    with mock.patch.dict(sys.modules, {'collections.abc': None})
        ...
```
Because `collections` is used also by numpy or pathlib which are imported during snap module imports.  

Do you have any idea on how to test this even in python3 ? 
if not we will need to  `# pragma: no cover`.

tomdele:
I will close this PR which is now obsolete with the python2 support removal from #109. I included the changes from this PR to the #109 one.
Thank you @tirkarthi for the contribution.

PR #91, opened 2020-09-14
Fix unit tests on Mac, fix doc indentation

GianlucaFicarelli:
Fix unit tests on Mac
- On Mac, tempfile.mkdtemp() returns something like `/var/folders/xxx` that's a simlink to `/private/var/folders/xxx`.
This PR resolves the symbolic link to allow the comparison of errors in unit tests.
- Set `pdf` as matplotlib backend in unit tests

Minor fix in docstring indentation

### DIFF
--- a/tests/test__plotting.py
+++ b/tests/test__plotting.py
@@ -11,6 +11,14 @@ def test__get_pyplot():
         with pytest.raises(ImportError):
             test_module._get_pyplot()
 
+    import matplotlib
+
+    # Set a backend different to MacOSX (default on Mac OS) to avoid the error
+    #   RuntimeError: Python is not installed as a framework
+    # using Python 2.7 in a virtualenv on Mac OS.
+    # See also https://matplotlib.org/faq/osx_framework.html.
+    matplotlib.use('pdf')
+
     import matplotlib.pyplot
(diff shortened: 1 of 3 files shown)


tests/test__plotting.py
```diff
@@ -11,6 +11,14 @@ def test__get_pyplot():
         with pytest.raises(ImportError):
             test_module._get_pyplot()
 
+    import matplotlib
+
+    # Set a backend different to MacOSX (default on Mac OS) to avoid the error
+    #   RuntimeError: Python is not installed as a framework
+    # using Python 2.7 in a virtualenv on Mac OS.
+    # See also https://matplotlib.org/faq/osx_framework.html.
+    matplotlib.use('pdf')
```
tomdele, line 20: I would more into :
```suggestion
    if sys.platform == 'darwin' and six.PY2:
        matplotlib.use('pdf')
```
so we can test the default one on linux ?

GianlucaFicarelli: Yes I think it's fine, we can change the matplotlib backend only in the known problematic case.

tomdele (approved):

tomdele:
thx !

PR #92, opened 2020-09-14
Improve _check_ids performance

GianlucaFicarelli:
Improve _check_ids performance.
See NSETM-1156 for investigation details.

### DIFF
--- a/bluepysnap/nodes.py
+++ b/bluepysnap/nodes.py
@@ -254,7 +254,7 @@ class NodePopulation(object):
 
     def _check_ids(self, node_ids):
         """Check that node IDs belong to the circuit."""
-        if not len(node_ids):
+        if len(node_ids) == 0:
             return
         # use the function with better performance for arrays or lists
         if isinstance(node_ids, np.ndarray):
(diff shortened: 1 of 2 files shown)


bluepysnap/nodes.py
```diff
@@ -254,9 +254,17 @@ def _check_id(self, node_id):
 
     def _check_ids(self, node_ids):
         """Check that node IDs belong to the circuit."""
-        missing = pd.Index(node_ids).difference(self._data.index)
-        if not missing.empty:
-            raise BluepySnapError("node ID not found: [%s]" % ",".join(map(str, missing)))
+        if not len(node_ids):
```
tomdele, line 257: I prefer to use : 
```suggestion
        if len(node_ids) == 0:
```

tests/test_nodes.py
```diff
@@ -255,6 +255,10 @@ def test_ids(self):
             _call('no-such-node-set')
         with pytest.raises(BluepySnapError):
             _call(-1)  # node ID out of range (lower boundary)
+        with pytest.raises(BluepySnapError):
+            _call([-1, 1])  # one of node IDs out of range (lower boundary)
+        with pytest.raises(BluepySnapError):
+            _call([1, -1])  # one of node IDs out of range, reversed order (lower boundary)
```
tomdele, line 261: good addition indeed

tomdele (approved):

tomdele:
Just as a general comment I (personally) don t like the 'bare not' too much because I think this is a bit too broad, error prone and reduce readability at the end.

PR #98, opened 2020-10-30
fix validation of required datasets of virtual node groups

asanin-epfl:
Some datasets are not required for virtual node groups. So we need to exclude virtual groups from most of node group checks.

### DIFF
--- a/bluepysnap/circuit_validation.py
+++ b/bluepysnap/circuit_validation.py
@@ -338,12 +338,14 @@ def _check_nodes_group(group_df, group, config):
     Returns:
         list: List of errors, empty if no errors
     """
-    REQUIRED_GROUP_NAMES = ['model_type', 'model_template']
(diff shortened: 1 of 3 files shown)


tomdele:
The coverage is failing. The code looks good otherwise.

tomdele (approved):

asanin-epfl:
ok, now I see. `h5py` library got updated to 3.0 and failed those tests. @tomdele how we proceed here? `h5py@3.0.0` seems too early for me to use. Currently it reads all our string fields in hdf5 files as binary => tests fail. Should be:
- `h5py` dependency be fixed for now as < 3.0
- expect and handle binary strings in bluepysnap?

tomdele:
Hum actually h5py is used only in the circuit validation otherwise everything is handled in libsonata.
I think you can force the version to < 3.0 for the moment. The impacts on moving to 3.0 are unclear to me right now. 
I am a bit scared for our other tools tho.

tomdele (approved):

PR #99, opened 2020-11-02
Add a circuit node interface

tomdele:
This pull request adds:
- A new class: CircuitNodeId 
  - A CircuitNodeIds is a combination of a population and a node id. --> ("population_name", node_id). 
- This implies to also allow the sequence of circuit node id as queries. So you have a low performance container using list of ids for the queries. We also provide a "high performance" one : CircuitNodeIds.
- A new class : CircuitNodeIds, based on pandas.MultiIndex to store and manipulate of global circuit ids. This is a high performance container for queries. I used composition and not heritage for this class because all the useful classmethods from pandas.MultiIndex return MutiIndex and not cls(). Therefore, implementing the `\_\_call\_\_` allows to use it as an index in the dataframes.

I chose this over a simple offsetting of ids because this is more compatible with the sonata main feature of "adding extra populations" to the circuit. With this design, the direct calls to ids in analysis scripts or the "on-disk stored" node ids (more common than we think ...) are not invalidated with new populations. We can also trustfully compare two sets of ids which is not the case with offsets. And since we allow people to use the sonata node ids or global ids this reduce the possible confusion and more flexibility for the API. I also don t want people to mix the sonata node ids and the global ids.
This also allows to do simple operations more transparently on ids like sorting/append/comparisons/filtering without relying on the circuit itself or mapping. 

Ex of uses for this class:
```
>>> ids_pop12 = CircuitNodeIds.from_arrays(["pop1", "pop1", "pop2"], [0, 1, 0])  # main 
[("pop1", 0), ("pop1", 1), ("pop2", 0)]
>>> ids_pop3 = CircuitNodeIds.from_dict({"pop3": [0, 1]})
[("pop3", 0), ("pop3", 1)]
ids_pop123 = ids_pop12.append(ids_pop3)
ids_pop1 = ids_pop123.filter_population("pop1")
sample_ids_pop123 = ids_pop123.sample(2)
```

- A new class: Nodes. This class aims at : 
  - Accessing the different node populations (like the current `circuit.nodes` does).
  - For compat you can iterate on the population using the dict interface : items(), values(), keys() or the wrapper populations(), names().
  - Allowing queries on the whole circuit for nodes. You can now do `circuit.nodes.ids(query)` and all the nodes from all populations matching the query will be returned. This should be handy for the spacial queries on different type of cells (like glia cells mixed with neurons) or for the combination of multiple regions merged together containing the same type of cells (like neocortex split in multiple sub region). 
  - Allowing queries on the whole circuit for nodes properties. You can now do `circuit.nodes.get(query, properties=[prop1, prop2])`. This will return a dataframe indexed by a MultiIndex (population, id). You will be able to query on this dataframe using : 
tuple ("population", id), CircuitNodeId, a list of tuples or a list of CircuitNodeId or a MultiIndex or CircuitNodeIds : 
```
ids = circuit.nodes.ids(query)  #  CircuitNodeIds corresponding to the query
data = circuit.nodes.get() # all data from all populations 
data.loc[ids]  # identical to circuit.nodes.get(query)
```
if a property is missing for a node id (this is possible due to population mixing) a NaN is attributed to the missing property value.

- The edges and morph class has been updated. You can use the CircuitNodeId/CircuitNodeIds inside the edges functions and the CircuitNodeId inside MorphHelper.
- Tests has been added accordingly.


tomdele:
The review is massive but it was hard to not forget something and the node ids are the lowest level of a sonata circuit so it lies pretty much everywhere ... 
If I can help you in whatever to do the review plus just ask me. 

An other review will follow to include this into the reports.

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+        ids = np.array([], dtype=np.int64)
+        populations = np.array([], dtype=str_type)
+        for name, pop in self.items():
+            pop_ids = pop.ids(group=group, raise_missing_property=False)
+            pops = np.array(np.full_like(pop_ids, fill_value=name, dtype=str_type))
+            ids = pop_ids if ids.size == 0 else np.concatenate([ids, pop_ids])
+            populations = pops if populations.size == 0 else np.concatenate([populations, pops])
+        return CircuitNodeIds.create_ids(populations, ids)
+
+    def get(self, group=None, properties=None):
```
asanin-epfl, line 197: Am I correct that if some properties are not presented, we will have a BluepySnapError? If yes, should we describe it in `Raises:`?

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+        if isinstance(group, CircuitNodeIds):
+            diff = np.setdiff1d(group.get_populations(unique=True), self.population_names)
+            if diff.size != 0:
+                raise BluepySnapError("Population {} does not exist in the circuit.".format(diff))
+        ids = np.array([], dtype=np.int64)
+        populations = np.array([], dtype=str_type)
+        for name, pop in self.items():
+            pop_ids = pop.ids(group=group, raise_missing_property=False)
+            pops = np.array(np.full_like(pop_ids, fill_value=name, dtype=str_type))
+            ids = pop_ids if ids.size == 0 else np.concatenate([ids, pop_ids])
```
asanin-epfl, line 193: Does it make sense to have this `if ids.size == 0`? It should work without it. The code will be easier to read. The same in the line below with `populations` .

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+        ids = self.ids(group)
+        if properties is None:
+            properties = self.property_names
+        properties = utils.ensure_list(properties)
+        res = pd.DataFrame(index=ids.index, columns=properties)
+        for name, pop in self.items():
+            global_pop_ids = ids.filter_population(name)
+            pop_ids = global_pop_ids.get_ids()
+            pop_properties = set(properties) & pop.property_names
+            for prop in pop_properties:
```
asanin-epfl, line 225: Why `for`? Can you put a comment why we can't set them at once?

asanin-epfl (commented):
Looks very good. I didn't finish yet because need to take a break for other tasks. Mostly typos notes.

tomdele:
First : thx a lot @asanin-epfl. I know the review is long so no problem if this takes time.
I will respond to your questions and wait for other's review.

tests/utils.py
```diff
@@ -97,8 +97,11 @@ def create_node_population(filepath, pop_name, circuit=None, node_sets=None):
     if node_sets is not None:
         circuit.node_sets = node_sets
     node_pop = NodeStorage(config, circuit).population(pop_name)
+
     if isinstance(circuit.nodes, dict):
         circuit.nodes[node_pop.name] = node_pop
```
asanin-epfl, line 102: Is this `if` branch actual? `circuit.nodes` now always Nodes, no?

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,216 @@
...
+            inplace (bool): if set to True. Do the transformation inplace.
+
+        Notes:
+            sorting a CircuitNodeIds will gives better perf for the population extraction and the
+            data extraction from a sorted dataframe.
+        """
+        if not inplace:
+            return CircuitNodeIds(self.index, sort_index=True)
+        self.index = self.index.sortlevel()[0]
+        return None
```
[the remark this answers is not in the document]
tomdele: It principle : no. But lint was not happy and they recommend to always return a value and in case of inplace like this to return None. 
So, I just followed the recommendations :
https://pycodequ.al/docs/pylint-messages/r1710-inconsistent-return-statements.html

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,216 @@
...
+            inplace (bool): if set to True. Do the transformation inplace.
+        """
+        res = self if inplace else self.copy()
+        res.index = res.index[0: limit_size]
+        if not inplace:
+            return res
+        return None
+
+    def to_csv(self, filepath):
+        """Save NodeCircuitIds to csv format."""
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,216 @@
...
+            sorting a CircuitNodeIds will gives better perf for the population extraction and the
+            data extraction from a sorted dataframe.
+        """
+        if not inplace:
+            return CircuitNodeIds(self.index, sort_index=True)
+        self.index = self.index.sortlevel()[0]
+        return None
+
+    def append(self, other, inplace=False):
+        """Append a NodeCircuitIds to the current one.
```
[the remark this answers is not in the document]
tomdele: Thx. That s actually a point I wanted to discuss should I call the class : NodeCircuitIds or CircuitNodeIds ? 
I did some back and forth about it and that s why you have a lot of typos ...

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,216 @@
...
+            return res
+        return None
+
+    def to_csv(self, filepath):
+        """Save NodeCircuitIds to csv format."""
+        self.index.to_frame(index=False).to_csv(filepath, index=False)
+
+    @classmethod
+    def from_csv(cls, filepath):
+        """Load NodeCircuitIds from csv."""
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/morph.py
```diff
@@ -55,23 +57,39 @@ def __init__(self, morph_dir, nodes):
...
     def get_filepath(self, node_id):
-        """Return path to SWC morphology file corresponding to `node_id`."""
-        name = self._nodes.get(node_id, Node.MORPHOLOGY)
+        """Return path to SWC morphology file corresponding to `node_id`.
+
+        Args:
+            node_id (int/CircuitNodeIds): could be a single int or a CircuitNodeIds of length 1.
+        """
+        if not isinstance(node_id, (CircuitNodeIds, np.integer, int)):
+            raise BluepySnapError("node_id must be either a single node id or a NodeCircuitIds "
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+            >>> # returns CircuitNodeIds matching all the queries inside the 'and' list
+            >>> nodes.ids(group={'$and': [{ Node.LAYER: [2, 3]},
+            >>>                           { Node.X: (0, 1), Node.MTYPE: 'L1_SLAC' }]})
+        """
+        str_type = "<U{}".format(max(len(pop) for pop in self.population_names))
+        if isinstance(group, CircuitNodeIds):
+            diff = np.setdiff1d(group.get_populations(unique=True), self.population_names)
+            if diff.size != 0:
+                raise BluepySnapError("Population {} does not exist in the circuit.".format(diff))
+        ids = np.array([], dtype=np.int64)
```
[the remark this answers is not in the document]
tomdele: Indeed. Changed.

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+        populations = np.array([], dtype=str_type)
+        for name, pop in self.items():
+            pop_ids = pop.ids(group=group, raise_missing_property=False)
+            pops = np.array(np.full_like(pop_ids, fill_value=name, dtype=str_type))
```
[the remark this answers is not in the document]
tomdele: That s a really good question ...
Removed it.

bluepysnap/nodes.py
tomdele: No we don t have a raise. But we should. I will add it. 
Actually the behavior should be : 
if the property is not in the circuit --> raise. 
if the property is just not in a some of the population --> fill with NaN.

I have added the check for existing property in the circuit, tests and the srt/list to the docstring.

bluepysnap/nodes.py
tomdele: Indeed you are right. Actually I remember I had problem with this before I added the  `raise_missing_property=False` due to incorrectly initialized pop_ids. 
I just did not rethink the whole function after adding the flag. 

I removed the if else this is much better.

bluepysnap/nodes.py
tomdele: Because not all properties are from the same type. So you cannot use a numpy array to set them directly. 
If you use a dataframe then it checks the indices and try to match them. But here they are different for the NodePopulation and the Nodes get functions.

bluepysnap/nodes.py
asanin-epfl: thank you

tests/utils.py
tomdele: Good eye, you are right !
I kept it for the edge tests but I changed stuffs in the edge tests and can remove it.

bluepysnap/edges.py
```diff
@@ -270,37 +268,45 @@ def positions(self, edge_ids, side, kind):
         result.sort_index(axis=1, inplace=True)
         return result
 
-    def afferent_nodes(self, node_id, unique=True):
+    def afferent_nodes(self, target, unique=True):
```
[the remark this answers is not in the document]
tomdele: indeed ... this a/efferent convention is not really the best.

would : 
```
Afferent nodes are nodes projecting an outgoing edge to one of the target node.
Efferent nodes are nodes receiving an incoming edge from one of the ``source`` node.
```
be ok ?

tests/test_nodes.py
```diff
@@ -2,25 +2,278 @@
...
+        assert isinstance(values[0], test_module.NodePopulation)
+        assert values[0].name == 'default'
+
+        assert isinstance(values[1], test_module.NodePopulation)
+        assert values[1].name == 'default2'
+
+        assert list(self.test_obj.values()) == list(self.test_obj.populations())
+
+    def test_items(self):
+        items = list(self.test_obj.items())
```
asanin-epfl, line 58: What do we test here?

asanin-epfl:
I've finished.

tests/test_nodes.py
tomdele: forgot this test. Added : 
```
        keys, values = zip(*self.test_obj.items())
        assert keys == ('default', 'default2')
        assert isinstance(values[0], test_module.NodePopulation)
        assert values[0].name == 'default'

        assert isinstance(values[1], test_module.NodePopulation)
        assert values[1].name == 'default2'
```

bluepysnap/edges.py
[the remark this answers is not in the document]
asanin-epfl: Yes. Better than mine! Everytime I stumble upon afferent/efferent, I have to google/see Sonata specs.

tomdele:
>I've finished. 

Thanks a lot I have added your last comments.

asanin-epfl:
Thank you, the code looks very good. I will wait for a day or two. If no one steps in, I will re-review last commits and we can merge.

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,216 @@
...
+        res = self if inplace else self.copy()
+        if len(res) != 0:
+            sample_size = sample_size if sample_size < len(res) else len(res)
+            res.index = res.index[np.random.choice(len(res), size=sample_size, replace=False)]
+        if not inplace:
+            return res
+        return None
+
+    def limit(self, limit_size, inplace=False):
+        """Sample a CircuitNodeIds.
```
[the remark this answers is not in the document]
tomdele: Forgot this comment. Yes this is a wild cut and paste. 
I need to change it.

bluepysnap/edges.py
```diff
@@ -121,9 +121,7 @@ def size(self):
 
     def _nodes(self, population_name):
         """Returns the NodePopulation corresponding to population."""
-        result = self._edge_storage.circuit.nodes.get(population_name)
-        if result is None:
-            raise BluepySnapError("Undefined node population: '%s'" % population_name)
+        result = self._edge_storage.circuit.nodes[population_name]
```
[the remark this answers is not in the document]
tomdele: Yes the node __getitem__ will raise instead.

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+        """Returns all the properties present inside the circuit."""
+        return set(prop for pop in self.populations() for prop in pop.property_names)
+
+    def property_values(self, prop):
+        """Returns all the values for a given property."""
+        return set(value for pop in self.populations() if prop in pop.property_names for value in
+                   pop.property_values(prop))
+
+    def ids(self, group=None):
+        """Retuns the CircuitNodeIds corresponding to the nodes from ``group``.
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+                depends on the type of the ``group`` argument:
+                - ``CircuitNodeIds``: return the IDs in a CircuitNodeIds object if they belong to
+                    the circuit.
+                - ``int``: if the node ID is present in all populations, returns a CircuitNodeIds
+                    object containing the corresponding node ID for all populations.
+                - ``sequence``: if all the values contained in the sequence are present in all
+                    populations, returns a CircuitNodeIds object containing the corresponding node
+                    IDs for all populations.
+                - ``str``: use a node set name as input. Returns a CircuitNodeIds object containing
+                    nodes selectected by the node set.
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+                    properties filter.
+                - ``None``: return all node IDs of the circuit in a CircuitNodeIds object.
+
+        Returns:
+            CircuitNodeIds: returns a CircuitNodeIds containing all the node IDs and the
+                corresponding populations. All the explicitly requested IDs must be present inside
+                the circuit.
+
+        Raises:
+            BluepySnapError: if a population is explicitly required by the user (via CircuitNodeIds)
```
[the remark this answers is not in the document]
tomdele: Changed for :           

 BluepySnapError: when a population from a CircuitNodeIds is not present in the circuit.
 BluepySnapError: when an id query via a int, sequence, or CircuitNodeIds is not present
                in the circuit.

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+            BluepySnapError: if one or more ids are explicilty asked by the user using int,
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+            >>> nodes.ids(group={Node.LAYER: [2, 3]})  # returns CircuitNodeIds with layer in [2,3]
+            >>> nodes.ids(group={Node.X: (0, 1)})  # returns CircuitNodeIds with 0 < x < 1
+            >>> # returns CircuitNodeIds matching one of the queries inside the 'or' list
+            >>> nodes.ids(group={'$or': [{ Node.LAYER: [2, 3]},
+            >>>                          { Node.X: (0, 1), Node.MTYPE: 'L1_SLAC' }]})
+            >>> # returns CircuitNodeIds matching all the queries inside the 'and' list
+            >>> nodes.ids(group={'$and': [{ Node.LAYER: [2, 3]},
+            >>>                           { Node.X: (0, 1), Node.MTYPE: 'L1_SLAC' }]})
+        """
+        str_type = "<U{}".format(max(len(pop) for pop in self.population_names))
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/nodes.py
```diff
@@ -32,12 +32,201 @@
...
+                the circuit.
+
+        Raises:
+            BluepySnapError: if a population is explicitly required by the user (via CircuitNodeIds)
+            BluepySnapError: if one or more ids are explicilty asked by the user using int,
+                sequence, or CircuitNodeIds.
+
+        Examples:
+            The available group parameter values (example with 2 nodes populations pop1 and pop2):
+            >>> nodes.ids(group=None)  #  returns all CircuitNodeIds from the circuit
```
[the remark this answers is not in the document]
tomdele: nodes is circuit.nodes with 2 populations "pop1" and "pop2"
Added to the example.

tomdele:
Sorry Aleksei the GH UI tricked me again and I missed some of your comments. 
I think I addressed everything now ...

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,222 @@
...
+
+        Args:
+            sample_size (int): the size of the sample. If the size of the sample is greater than
+                the size of the CircuitNodeIds then all ids are taken and shuffled.
+            inplace (bool): if set to True. Do the transformation inplace.
+        """
+        res = self if inplace else self.copy()
+        if len(res) != 0:
+            sample_size = sample_size if sample_size < len(res) else len(res)
+            res.index = res.index[np.random.choice(len(res), size=sample_size, replace=False)]
```
wizmer, line 174:
```suggestion

            res.index = res.index[np.random.choice(len(res), size=min(sample_size, len(res)), replace=False)]
```

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,222 @@
...
+        if not inplace:
+            return res
+        return None
+
+    def limit(self, limit_size, inplace=False):
```
wizmer, line 179: What about `resize` instead of `limit` ? Or `shrink`.
IMO what is missing in `limit` is the fact we are talking about size capacity. And not limit in the values themselves for example.

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,222 @@
...
+        Notes:
+            this function takes the first ``limit_size`` values of a circuit node ids.
+
+        Args:
+            limit_size (int): the size of the sample. If the size of the sample is greater than
+                the size of the CircuitNodeIds then all ids are kept.
+            inplace (bool): if set to True. Do the transformation inplace.
+        """
+        res = self if inplace else self.copy()
+        res.index = res.index[0: limit_size]
```
wizmer, line 191: `limit` and `sample` are the same code except for the slice passed in `res.index[]`.

It may not be useful now, but for the future you can consider having a private function:

```python
def _slice_index(self, myslice, inplace=False):
        res = self if inplace else self.copy()
        res.index = res.index[myslice]
        if not inplace:
            return res
        return None
```

Then:
```python
     def limit(self, limit_size, inplace=False):
         return self._slice_index([0, limit_size], inplace=inplace)

     def sample(self, limit_size, inplace=False):
         # this one does not handle the len(res) == 0 but you see the idea
         return self._slice_index(np.random.choice(len(res), size=min(sample_size, len(res)), replace=False), 
                                  inplace=inplace)
```

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,222 @@
...
+        """Correct str of the CircuitNodeIds."""
+        return self.__repr__()
+
+    def __eq__(self, other):
+        """Equality for the CircuitNodeIds."""
+        if not isinstance(other, CircuitNodeIds):
+            return False
+        return self.index.equals(other.index)
+
+    def __len__(self):
```
wizmer, line 220: Have you tried implementing directly \_\_iter\_\_ instead of \_\_len\_\_ ?

bluepysnap/edges.py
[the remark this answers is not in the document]
wizmer: Hello my name is Thomas and I love __bold__ because I like underscores

bluepysnap/circuit_ids.py
wizmer: Maybe this comment was not clear.
I *suppose* that if you do:

```python
def __iter__(self):
    return iter(self.index)
```
then len(self) should be len(self.index) automatically ?

bluepysnap/morph.py
```diff
@@ -55,23 +57,39 @@ def __init__(self, morph_dir, nodes):
...
-        return self._nodes.get(node_id, Node.MODEL_TYPE) == "biophysical"
+        return self._population.get(node_id, Node.MODEL_TYPE) == "biophysical"
 
     def get_filepath(self, node_id):
-        """Return path to SWC morphology file corresponding to `node_id`."""
-        name = self._nodes.get(node_id, Node.MORPHOLOGY)
+        """Return path to SWC morphology file corresponding to `node_id`.
+
+        Args:
+            node_id (int/CircuitNodeIds): could be a single int or a CircuitNodeIds of length 1.
```
wizmer, line 66: I really don't like it. int and CircuitNodeIds are not homogeneous. One is a scalar and the other one is homogeneous to a sequence. 

I would prefer having an [] accessor in the CircuitNodeIds class and having support of the following:
```python
morph_helper.get_filepath(circuit_node_ids[0])
```
This support would also enable this:
```python
filepaths = [morph_helper.filepath(node_id) for node_id in circuit_node_ids]
```

bluepysnap/morph.py
```diff
@@ -55,23 +57,39 @@ def __init__(self, morph_dir, nodes):
...
+                                  "with a single value. You provided : {}.".format(node_id))
+        if isinstance(node_id, CircuitNodeIds) and len(node_id) > 1:
+            raise BluepySnapError("node_id must be either a single node id or a CircuitNodeIds "
+                                  "with a single value. Length is {}.".format(len(node_id)))
+
+        name = self._population.get(node_id, Node.MORPHOLOGY)
+        # to allow the NodeCircuitIds as node_id
+        if not isinstance(name, six.string_types):
+            name = name.squeeze()
         return os.path.join(self._morph_dir, "%s.swc" % name)
```
wizmer, line 79: Should not this return a Path ? Maybe for another PR.

bluepysnap/nodes.py
```diff
@@ -436,8 +649,10 @@ def ids(self, group=None, limit=None, sample=None):
             result = utils.ensure_list(group)
             self._check_ids(result)
             preserve_order = isinstance(group, collections.Sequence)
+
         if sample is not None:
             if len(result) > 0:
+                sample = sample if sample <= len(result) else len(result)
```
wizmer, line 655: You don't like `min` don't you ? :)

bluepysnap/nodes.py
```diff
@@ -394,10 +599,15 @@ def ids(self, group=None, limit=None, sample=None):
...
             limit (int): If specified, return the first ``limit`` number of
-                IDs from the match result.
+                IDs from the match result. If limit is greater than the size of the population
+                all node IDs are returned.
+
+            raise_missing_property (bool): if True, raises if a property is not listed in this
+                population. Otherwise the ids are just not selected if a property is missing.
 
         Returns:
             numpy.array: A numpy array of IDs.
```
wizmer, line 613: You should probably add an example with circuitnodeids in the example list below

bluepysnap/nodes.py
```diff
@@ -453,22 +668,24 @@ def get(self, group=None, properties=None):
...
 
                 - ``int``: return the properties of a single node.
+                - ``CircuitNodeIds`` return the properties from a NodeCircuitNodeIds.
                 - ``sequence``: return the properties from a list of node.
                 - ``str``: return the properties of nodes in a node set.
                 - ``mapping``: return the properties of nodes matching a properties filter.
                 - ``None``: return the properties of all nodes.
 
-            properties (set): If specified, return only the properties in the set.
+            properties (list): If specified, return only the properties in the list.
```
wizmer, line 681: why not all sequences ?

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,222 @@
...
+import numpy as np
+import pandas as pd
+import six
+
+from bluepysnap import utils
+from bluepysnap.exceptions import BluepySnapError
+
+
+class CircuitNodeIds:
+    """Global Node ids.
```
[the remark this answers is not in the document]
asanin-epfl: CircuitNodeIds os fine for me. NodeIndex is too broad.

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,222 @@
...
+        if not isinstance(index, pd.MultiIndex):
+            raise BluepySnapError("index must be a pandas.MultiIndex object.")
+        index.names = ["population", "node_ids"]
+        if sort_index:
+            # best perf compared to sort_values. Sorted by population and ids.
+            index = index.sortlevel()[0]
+        self.index = index
+
+    @classmethod
+    def create_ids(cls, populations, population_ids, sort_index=True):
```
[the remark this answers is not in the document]
tomdele: I changed the api:

- I renamed create_ids in from_array.
- I removed the possibility to from_array("population", [1,2]) now to do this you need to use from_dict as you proposed
- I added a from_tuples
- I added a from_dict

bluepysnap/circuit_ids.py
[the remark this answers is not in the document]
tomdele: I don t want "global' because this is too close to the gids from old format in bluepy
Index only is too broad because you can have "edge index" also.
I think I will stick to CircuitNodeIds/CircuitNodeId

bluepysnap/circuit_ids.py
tomdele: done

bluepysnap/circuit_ids.py
tomdele: I will stick to limit because it refers to the node API in NodePopulation: 
```
def ids(self, group=None, limit=None, sample=None)
```

If I change this I will change it for all the NodePopulation also.

bluepysnap/circuit_ids.py
tomdele: indeed. I totally forgot to factorize. 
done

bluepysnap/circuit_ids.py
tomdele: I have a \_\_iter\_\_ and a \_\_len\_\_ now.

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @classmethod
+    def from_csv(cls, filepath):
+        """Load CircuitNodeIds from csv."""
+        return cls(pd.MultiIndex.from_frame(pd.read_csv(filepath)))
+
+    def __repr__(self):
+        """Correct repr of the CircuitNodeIds."""
+        res = self.index.__repr__()[len("MultiIndex"):]
+        return "CircuitNodeIds" + res
```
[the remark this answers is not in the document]
tomdele: I think mine has better perf ...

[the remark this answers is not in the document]
tomdele: I changed it anyway to please you ...

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __str__(self):
+        """Correct str of the CircuitNodeIds."""
+        return self.__repr__()
```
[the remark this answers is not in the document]
tomdele: indeed. I removed it.

bluepysnap/circuit_ids.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __eq__(self, other):
+        """Equality for the CircuitNodeIds."""
+        if not isinstance(other, CircuitNodeIds):
```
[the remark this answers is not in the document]
tomdele: yeah why not I did it.

bluepysnap/morph.py
```diff
@@ -19,29 +19,31 @@
 
 import os
 
+import six
```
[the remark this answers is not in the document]
tomdele: _wizmer receives bluepy power from tomdele_

bluepysnap/morph.py
tomdele: I changed this for CircuitNodeId both are homogeneous to a single id

bluepysnap/morph.py
tomdele: nope because python2

bluepysnap/nodes.py
tomdele: lol this is my classic tunnel vision when I code

bluepysnap/nodes.py
tomdele: done

bluepysnap/nodes.py
tomdele: because we call the properties directly from the dataframe so tuple fails.

tomdele:
I think I made all your requested benoit. 
I also added the CircuitNodeId which makes the api more complete and thanks to @wizmer  I found a way to use the circuit node ids directly as index in the dataframes thanks to the \_\_call\_\_ (you see @wizmer I can do underscore now).

Thanks a lot @wizmer for the help.

tomdele:
And we have a CI again :')

asanin-epfl (approved):
Great job, Thomas!

tomdele:
Thanks a lot both of you for the very valuable comments !

PR #97, opened 2020-10-13
Kgforge (nexus-forge)

joni-herttuainen:
Adding the updated script to BlueBrain's repo:

BluePy support (for demoing purposes only)
Cross-Bucket search/retrieve
Jupyter notebook for kgforge stuff

PR #65, opened 2020-06-23
Added kgforge.py (KGForge wrapper for Simulation and Circuit)

joni-herttuainen:

PR #107, opened 2020-11-17
Kgforge (nexus-forge) (#97)

joni-herttuainen:
* Added kgforge.py (KGForge wrapper for Simulation and Circuit) (#65)

* Updated kgforge.py (Cross-bucket search/retrieve and BluePy support for demoing)

* Added notebook for kgforge (nexusforge)

**NOTE:**
* nexusforge(kgforge) or bluepy are not in the setup.py and these need to be installed manually to run kgforge.py
* bluepy will never be supported, it is here only for demoing purposes

### DIFF
--- a/bluepysnap/kgforge.py
+++ b/bluepysnap/kgforge.py
@@ -1,0 +1,317 @@
+"""KnowledgeGraphForge wrappings for SNAP.
+
+General notes:
+    User logging in runtime not yet implemented
+    - Did some tests, and there seems to be an issue regarding the Oauth2
+      client secret which we do not have.
+    - Should be investigated further.
+    - So for now, need to copy the token for the kgforge
+
+    Getting resources created by current user not implemented as of yet
+    - Would be simple, if we had a Oauth2 login and could get the username from there
+
+    Simulation not tested yet due to there not being Simulations with SONATA config in Nexus
+    - SNAP does not support BlueConfig
+
+    Currently the functions retrieve all the instances in the KG, also the deprecated ones
+    - if this is not wanted, there is a way to exclude them in the sparql queries
+
+    KnowledgeGraphForge instance is passed to the classes' __init__ function
+    - I would prefer a global instance of it but then it would need to be changed if accessing
+      data in other 'buckets' (org/project)
+    - not sure what is the best approach for this or if this is going to be simpler in the future.
+"""
+
+
+import os
+import logging
+from cached_property import cached_property
+
+# from bluepysnap import Circuit, Simulation
+from bluepysnap import Circuit
+from bluepy.v2 import Simulation as BluepySimulation, Circuit as BluepyCircuit
+
+L = logging.getLogger(__name__)
+
+
+def _statement_deprecated(deprecated):
+    """Helper function to return a query statement for deprecated property."""
+    keyword = 'https://bluebrain.github.io/nexus/vocabulary/deprecated'
+    bool_str = str(bool(deprecated)).lower()
+    return '?id <{}> {} .'.format(keyword, bool_str)
+
+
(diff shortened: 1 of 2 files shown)


tomdele:
lol KG Forge is not python2 compatible

wizmer:
:popcorn:

joni-herttuainen:
> lol KG Forge is not python2 compatible

True. Didn't remember that one. I guess snap won't be, either, for long?

tomdele:
I have no idea ...

tomdele:
The python2 support will be drop here :  #109

tomdele:
#109 has been merged

bluepysnap/kgforge.py
```diff
@@ -0,0 +1,317 @@
...
+"""
+
+
+import os
+import logging
+from cached_property import cached_property
+
+# from bluepysnap import Circuit, Simulation
+from bluepysnap import Circuit
+from bluepy.v2 import Simulation as BluepySimulation, Circuit as BluepyCircuit
```
tomdele, line 32: I am not gonna lie this is really strange having this here...

asanin-epfl: Can it even be merged here?

tomdele: And we have a potential cyclic dependency because snap is a dep of bluepy and bluepy is imported in snap.

tomdele: > Can it even be merged here?

Not sure about the linting. Maybe if the import is lazy.

asanin-epfl: I would avoid bluepy here at all.

tomdele: definitely ...

tomdele: I think @mgeplf this is your call here. 
To me the whole kgforge thing should be in another package to be honest. But I will follow your call.

joni-herttuainen: I have no objection to this being its own package. 
Regarding Bluepy, I was under the impression that it will not stay here.

tomdele:
Actually stupid call but why snap is using kgforge and not the other way around ? 
why KGForge cannot have something to retrieve snap circuit instead ? It makes more sense to me...

joni-herttuainen:
> Actually stupid call but why snap is using kgforge and not the other way around ?
> why KGForge cannot have something to retrieve snap circuit instead ? It makes more sense to me...

kgforge is an API to retrieve data and metadata of instances registered in Nexus. Not all the data used by Snap is in Nexus, so I'm not convinced that is the best approach, either.

That being said, I am not convinced that having the kgforge wrappers in Snap is the best approach either:  It is a tool to access Circuit  and Simulation data that is in Sonata format and technically, my code does nothing of the sort. It just uses the combines kgforge and snap. 

The more I think about it, I feel this should be a separate package.

bluepysnap/kgforge.py
mgeplf: Yeah, we can't really have bluepy refs in here; can we make an extension point that bluepy can hook to make this work?  I'd prefer it not be a monkey patch situation, for cleanliness sake.

mgeplf:
It's been a requirement from @jdcourcol that it's part of snap; I also suspect it needs to be in a separate package, but at this point we can hopefully continue with it being in one package until it's untennable.

For it to work as another package, we'll have to add the extension points as I mentioned above - although I don't know where they all are - so I consider it a good chance to see where they are, and thus show it's feasible that it's another package.

tomdele:
Ok ! 
Thanks for the clarification @mgeplf. I did not follow the full story and needed some insights here.

PR #108, opened 2020-11-19
Allow usage of config dict instead of file only

tomdele:
* this allows consumer codes to use a different input file config
  format and still being able to convert it into dict and pass it to the
  Circuit/Simulation classes.


setup.py
```diff
@@ -44,7 +44,8 @@ def __init__(self, *args, **kwargs):
         'cached_property>=1.0',
         'functools32;python_version<"3.2"',
         'h5py>=2.2,<3',
-        'libsonata>=0.1.4',
+        'libsonata<=0.1.4;python_version<"3.6"',
```
tomdele, line 47: I hope we will drop this soon enough. 
libsonata api did not change and the result are identical. Python2 users will just not benefit from the perf improvement

tests/test_simulation.py
```diff
@@ -80,3 +81,32 @@ def test_no_node_set():
...
+    with pytest.raises(BluepySnapError):
+        test_module.Simulation(input_dict)
+
+    input_dict = {
+        "network": str(TEST_DATA_DIR / "./circuit_config.json"),
+        "simulation": str(TEST_DATA_DIR / "./simulation_config.json")
+    }
+    simulation = test_module.Simulation(input_dict)
+    assert simulation.config["network"] == str(TEST_DATA_DIR / 'circuit_config.json')
+    assert sorted(list(simulation.circuit.nodes)) == ['default', 'default2']
```
wizmer, line 100: Eventually:

```suggestion
    assert set(simulation.circuit.nodes) == {'default', 'default2'}
```

bluepysnap/config.py
```diff
@@ -35,29 +35,32 @@ class Config(object):
         https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#network_config
     """
 
-    def __init__(self, filepath):
+    def __init__(self, config):
         """Initializes a Config object from a path to the actual config.
 
         Args:
-            filepath (str): Path the SONATA configuration file.
+            config (str/dict): Path to the SONATA configuration file or dict containing the config.
```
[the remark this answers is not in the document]
tomdele: Actually this is on purpose. Doing it here allow to factorize the `if/else` instead of delegating it to all the consumer class. 
So basically not having the `if/else` here would mean to have exactly the same code in all the consumers: 
```
    if isinstance(config, dict):
         self._config = Config.load_dict().resolve()
    elif isinstance(config, ...):
         ....
```
Which is not optimal imao.

bluepysnap/config.py
```diff
@@ -35,29 +35,32 @@ class Config(object):
...
 
     @staticmethod
     def _resolve_manifest(manifest, configdir):
         result = manifest.copy()
 
-        assert '${configdir}' not in result
-        result['${configdir}'] = configdir
-
         for k, v in six.iteritems(result):
             if v.startswith('.'):
```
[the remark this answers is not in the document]
tomdele: good point I wanted to change this for is_absolute but forgot about it !
I expect only path like string in the manifest so this is doable indeed.

bluepysnap/config.py
```diff
@@ -79,6 +82,9 @@ def _resolve_manifest(manifest, configdir):
             if not v.startswith('/'):
                 raise BluepySnapError("{} cannot be resolved as an abs path.".format(k))
 
+        assert '${configdir}' not in result
```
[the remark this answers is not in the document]
tomdele: I am not sure what you mean here. 
'${configdir}' can be none here because if you use a dict instead of a file you cannot define a configdir. Otherwise it must exists because this is the input path of the config (if it is a file)

bluepysnap/config.py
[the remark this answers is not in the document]
tomdele: Ah I remember why know ... 
I need to be careful with the string starting with $which will be detected a relative path and resolved...

bluepysnap/config.py
```diff
@@ -79,6 +82,9 @@ def _resolve_manifest(manifest, configdir):
...
+        result['${configdir}'] = configdir
+
         return result
 
     def _resolve_string(self, value):
```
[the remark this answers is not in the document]
tomdele: If the path sub directory starts with a `$` it fails. If the sub directory only include $ but not at the starting positions it does not fail. 
```
"/something/$UNKNOWN/" --> fails 
"/something/U$N$K$N$O$W$N/" --> works
```

tests/test_config.py
```diff
@@ -123,3 +124,26 @@ def test_simulation_config():
     assert actual["mechanisms_dir"] == str(Path(TEST_DATA_DIR / "../shared_components_mechanisms").resolve())
     assert actual["conditions"]["celsius"] == 34.0
     assert actual["conditions"]["v_init"] == -80
+
+
+def test_dict_config():
```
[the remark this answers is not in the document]
tomdele: It will fail for the moment. When we dropped python2 will make it transparent for both.

tests/test_config.py
```diff
@@ -123,3 +124,26 @@ def test_simulation_config():
...
+    config = json.load(open(str(TEST_DATA_DIR / 'circuit_config.json'), "r"))
+    with pytest.raises(BluepySnapError):
```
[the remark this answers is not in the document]
tomdele: sure

tests/test_simulation.py
tomdele: yes changed it

bluepysnap/config.py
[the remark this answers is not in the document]
wizmer: ok nevermind

tests/test_simulation.py
```diff
@@ -14,7 +14,7 @@
 def test_all():
     simulation = test_module.Simulation(str(TEST_DATA_DIR / 'simulation_config.json'))
     assert simulation.config["network"] == str(TEST_DATA_DIR / 'circuit_config.json')
-    assert sorted(list(simulation.circuit.nodes)) == ['default', 'default2']
+    assert set(list(simulation.circuit.nodes)) == set(['default', 'default2'])
```
wizmer, line 17:
```suggestion
    assert set(simulation.circuit.nodes) == set(['default', 'default2'])
```

tests/test_simulation.py
```diff
@@ -31,16 +31,16 @@ def test_all():
     assert isinstance(simulation.spikes, SpikeReport)
     assert isinstance(simulation.spikes["default"], PopulationSpikeReport)
 
-    assert sorted(list(simulation.reports)) == sorted(list(['soma_report', 'section_report']))
+    assert set(list(simulation.reports)) == set(list(['soma_report', 'section_report']))
```
wizmer, line 34:
```suggestion
    assert set(simulation.reports) == {'soma_report', 'section_report'}
```

tests/test_simulation.py
```diff
@@ -31,16 +31,16 @@ def test_all():
...
     assert isinstance(simulation.reports['soma_report'], SomaReport)
     assert isinstance(simulation.reports['section_report'], CompartmentReport)
 
     rep = simulation.reports['soma_report']
-    assert sorted(list(rep.population_names)) == ["default", "default2"]
+    assert set(list(rep.population_names)) == {"default", "default2"}
```
wizmer, line 39:
```suggestion
    assert set(rep.population_names) == {"default", "default2"}
```

tests/test_simulation.py
```diff
@@ -31,16 +31,16 @@ def test_all():
...
     assert isinstance(rep['default'], PopulationSomaReport)
 
     rep = simulation.reports['section_report']
-    assert sorted(list(rep.population_names)) == ["default", "default2"]
+    assert set(list(rep.population_names)) == {"default", "default2"}
```
wizmer, line 43:
```suggestion
    assert set(rep.population_names) == {"default", "default2"}
```

tests/test_simulation.py
```diff
@@ -61,11 +61,11 @@ def test_unknown_report():
 def test__resolve_config():
     simulation = test_module.Simulation(str(TEST_DATA_DIR / 'config.json'))
     assert simulation.config["network"] == str(TEST_DATA_DIR / 'circuit_config.json')
-    assert sorted(list(simulation.circuit.nodes)) == ['default', 'default2']
+    assert set(list(simulation.circuit.nodes)) == set(['default', 'default2'])
```
wizmer, line 64:
```suggestion
    assert set(simulation.circuit.nodes) == {'default', 'default2'}
```

asanin-epfl:
I had a quick glance. Looks fine for me.

tests/test_simulation.py
tomdele: lol I think I was at "le bout de ma vie".

wizmer (approved):

wizmer:
ggwp

tomdele:
gogog

Thank you guys !

PR #110, opened 2020-12-01
Remove six dependency

tomdele:
Removing the six dependency for tests and codes


joni-herttuainen (approved):
Looks good to me.

tomdele:
Thank you @joni-herttuainen !

PR #113, opened 2020-12-11
Factorization for Edges and Nodes

tomdele:
* Add NetworkObject
* Add AbstractDocSubstitutionMeta the Edges/Nodes class
* Add get to the Edges to match the node interface
* Deprecate the returning of ids using get/properties for edges

### DIFF
--- a/bluepysnap/_doctools.py
+++ b/bluepysnap/_doctools.py
@@ -18,6 +18,7 @@
 import inspect
 import types
 import functools
+import abc
 
 
 def _word_swapper(doc, source_word, target_word):
@@ -38,20 +39,34 @@ def _copy_func(f):
 
 # better than decorator to do that due to the returned type being correct with this
 # with wrapper <class 'bluepysnap._doctools.DocSubstitutionDecorator.__call__.<locals>.Wrapped'>
-# works well with Sphinx also
+# works well with Sphinx also.
 class DocSubstitutionMeta(type):
     """Tool to update an inherited class documentation."""
     def __new__(mcs, name, parents, attrs, source_word=None, target_word=None):
         """Define the new class to return."""
         for parent in parents:
+            # skip classmethod with isfunction if I use also ismethod as a predicate I can have the
+            # classmethod docstring changed but then the cls argument is not automatically skipped.
             for fun_name, fun_value in inspect.getmembers(parent, predicate=inspect.isfunction):
+                # skip abstract methods. This is fine because we must override them anyway
+                try:
+                    if fun_name in parent.__abstractmethods__:
+                        continue
+                except AttributeError:
+                    pass
                 # skip special methods
                 if fun_name.startswith("__"):
                     continue
-
                 changed_fun = _copy_func(fun_value)
                 changed_fun.__doc__ = _word_swapper(changed_fun.__doc__, source_word, target_word)
                 attrs[fun_name] = changed_fun
         # create the class
         obj = super(DocSubstitutionMeta, mcs).__new__(mcs, name, parents, attrs)
         return obj
+
+
+class AbstractDocSubstitutionMeta(abc.ABCMeta, DocSubstitutionMeta):
+    """Mixin class to use with abstract classes.
+
+    It solves the metaclass conflict.
+    """
--- a/bluepysnap/_doctools.py
+++ b/bluepysnap/_doctools.py
@@ -37,11 +37,19 @@ def _copy_func(f):
     return g
 
 
-# better than decorator to do that due to the returned type being correct with this
-# with wrapper <class 'bluepysnap._doctools.DocSubstitutionDecorator.__call__.<locals>.Wrapped'>
-# works well with Sphinx also.
 class DocSubstitutionMeta(type):
-    """Tool to update an inherited class documentation."""
+    """Tool to update an inherited class documentation.
+
+    Notes:
+        Using a metaclass is better than decorator to do that due to the returned type being
+        incorrect when using a wrapper. Ex with CircuitNodeIds:
+        type(CircuitNodeIds)
+        <class 'bluepysnap._doctools.DocSubstitutionDecorator.__call__.<locals>.Wrapped'>
+        when with metaclass:
+        type(CircuitNodeIds)
+        <class 'bluepysnap.CircuitNodeIds'>
+        It works well with Sphinx also.
+    """
     def __new__(mcs, name, parents, attrs, source_word=None, target_word=None):
         """Define the new class to return."""
         for parent in parents:
--- a/bluepysnap/edges.py
+++ b/bluepysnap/edges.py
@@ -366,13 +366,6 @@ class EdgeStorage:
         return self._populations[population_name]
 
 
-def _resolve_node_ids(nodes, group):
-    """Node IDs corresponding to node group filter."""
-    if group is None:
-        return None
-    return nodes.ids(group)
-
-
 def _is_empty(xs):
     return (xs is not None) and (len(xs) == 0)
 
@@ -406,6 +399,13 @@ class EdgePopulation:
     @cached_property
     def _population(self):
         return self._edge_storage.storage.open_population(self.name)
+
+    @staticmethod
+    def _resolve_node_ids(nodes, group):
+        """Node IDs corresponding to node group filter."""
+        if group is None:
+            return None
+        return nodes.ids(group)
 
     @property
     def size(self):
@@ -616,7 +616,7 @@ class EdgePopulation:
         """
         if target is not None:
             selection = self._population.afferent_edges(
-                _resolve_node_ids(self.target, target)
+                self._resolve_node_ids(self.target, target)
             )
         else:
             selection = self._population.select_all()
@@ -641,7 +641,7 @@ class EdgePopulation:
         """
         if source is not None:
             selection = self._population.efferent_edges(
-                _resolve_node_ids(self.source, source)
+                self._resolve_node_ids(self.source, source)
             )
         else:
             selection = self._population.select_all()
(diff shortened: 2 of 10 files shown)


bluepysnap/edges.py
```diff
@@ -24,94 +24,34 @@
...
 
-    def __init__(self, circuit):
+    def __init__(self, circuit):  # pylint: disable=useless-super-delegation
         """Initialize the top level Edges accessor."""
-        self._circuit = circuit
-        self._config = self._circuit.config['networks']['edges']
-        self._populations = self._get_populations()
+        super().__init__(circuit)
 
     def _get_populations(self):
```
asanin-epfl, line 44: I really hoped that `_get_populations`,`ids`,`get` would be factorized

bluepysnap/edges.py
```diff
@@ -506,7 +456,7 @@ def property_dtypes(self):
         Returns:
             pandas.Series: series indexed by field name with the corresponding dtype as value.
         """
-        return self.properties([0], list(self.property_names)).dtypes.sort_index()
+        return self.get([0], list(self.property_names)).dtypes.sort_index()
 
     def container_property_names(self, container):
         """Lists the ConstContainer properties shared with the EdgePopulation.
```
asanin-epfl, line 462: At the line 475 there is a call to `my_edge_population.property_values`. Is there such method of EdgePopulation?

bluepysnap/edges.py
```diff
@@ -135,7 +75,7 @@ def _get_ids_from_pop(self, fun_to_apply, returned_ids_cls):
         populations = np.concatenate(populations).astype(str_type)
         return returned_ids_cls.from_arrays(populations, ids)
 
-    def ids(self, edge_ids):
+    def ids(self, edge_ids):  # pylint: disable=arguments-differ
```
asanin-epfl, line 78: `CircuitNodeIds` at the line 88

bluepysnap/edges.py
```diff
@@ -182,6 +122,8 @@ def properties(self, edge_ids, properties):
         ids = self.ids(edge_ids)
         # TODO : remove this due to the addition of ids to the EdgePopulation
```
asanin-epfl, line 123: Is this TODO actual?

bluepysnap/edges.py
tomdele: I must say I did not find a good way of doing it...

The problem are the specific classes used in these functions : EdgeStorage, CircuitEdgeIds, CircuitEdgeId.

This is indeed easy for the get.

bluepysnap/edges.py
tomdele: done

bluepysnap/edges.py
tomdele: Yes it is related to the Deprecate below.

bluepysnap/edges.py
tomdele: no !

bluepysnap/edges.py
tomdele: I made something but this is not quite perfect.

asanin-epfl (approved):
lgtm. One thought for discussion. What if we could keep only CircuitIds instead of CircuitNodeIds, CircuitEdgeIds? They are basically the same. The only substantial difference is the context where they used.

tomdele:
> The only substantial difference is the context where they used.

That's exactly the point. With only one class `CircuitIds`, a "CircuitNodeId" could work on a edge id based function if the population names are similar between nodes and edges (let say : population names = "All" ...).
I have the feeling that this can lead to problems afterwards. 

Also it prevents consumer functions to have : 
```
if isinstance(ids, CircuitNodeIds):
    ...
elif  isinstance(ids, CircuitEdgeIds):
    ...
```
Also imagine one day the implementations change and starts to differ for the edge ids and node ids. Then we will be in trouble. 

That was my thoughts when doing the separation. What do you think ?

tomdele:
@joni-herttuainen , @wizmer , @mgeplf, @asanin-epfl 
If you all think this is a bad idea to keep the ids split in node and edge ids, I am ok to change it.

asanin-epfl:
I don't think that this is a bad idea.

tomdele:
Thanks @asanin-epfl for the review !
To others : do you want to review ?

mgeplf:
I would like to read it, but time is a problem.

bluepysnap/network.py
```diff
@@ -0,0 +1,170 @@
...
+                if population in res:
+                    raise BluepySnapError(
+                        "Duplicated {} population: '{}'".format(self.__class__.__name__,
+                                                                population))
+                res[population] = storage.population(population)
+        return res
+
+    @abc.abstractmethod
+    def _collect_populations(self):
+        """Should specify the self._get_populations arguments."""
```
wizmer, line 51: You did not put a NotImplementedError on purpose here ?

joni-herttuainen:
Taking a look now

bluepysnap/network.py
```diff
@@ -0,0 +1,170 @@
...
+        return self._circuit.config
+
+    @cached_property
+    def _populations(self):
+        """Cached population dictionary."""
+        return self._collect_populations()
+
+    @cached_property
+    def population_names(self):
+        """Returns all the NetworkObjects population names from the Circuit."""
```
wizmer, line 64: You can add the fact it returns a sorted object.

bluepysnap/network.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @abc.abstractmethod
+    def ids(self, *args, **kwargs):
+        """Resolves the ids of the NetworkObject."""
+
+    @abc.abstractmethod
+    def get(self, group=None, properties=None):
+        """Returns the properties of a the NetworkObject."""
+        ids = self.ids(group)
+        properties = utils.ensure_list(properties)
```
wizmer, line 154: you could have a properties_set var here since you do set(properties) multiple times.

bluepysnap/nodes.py
```diff
@@ -27,103 +27,35 @@
...
+        return self._get_populations(NodeStorage, self._config['networks']['nodes'])
 
     def property_values(self, prop):
-        """Returns all the values for a given property."""
+        """Returns all the values for a given Nodes property."""
         return set(value for pop in self.values() if prop in pop.property_names for value in
                    pop.property_values(prop))
 
-    def ids(self, group=None):
+    def ids(self, group=None):   # pylint: disable=arguments-differ
```
wizmer, line 58: Just FYI: The rationale they give for this warning is pretty good.
https://github.com/PyCQA/pylint/issues/3638

bluepysnap/network.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+class NetworkObject(abc.ABC):
+    """Abstract class for the top level NetworkObjects accessor."""
+
+    def __init__(self, circuit):
+        """Initialize the top level NetworkObjects accessor."""
+        self._circuit = circuit
+
+    def _get_populations(self, cls, config):
```
wizmer, line 36: If I understand correctly, this function will always be called with cls=EdgeStorage when self is  of type Edges and cls=NodeStorage when self is of type Nodes.

In this case, why not having a class level attribute Storage in the NetworkObject class that you set to 
Storage=EdgeStorage in Edges
Storage=NodeStorage in Nodes

Then in the following code you can remove cls from the argument list and replace cls by Storage in the body.
Conceptually I think it also makes it more explicit that Edges is tied to EdgeStorage the same way Nodes is tied to NodeStorage.

joni-herttuainen (commented):
All in all, looks good to me. Also, I think that having CircuitNodeIds and CircuitEdgeIds separate is justified.

bluepysnap/_doctools.py
```diff
@@ -38,20 +39,34 @@ def _copy_func(f):
 
 # better than decorator to do that due to the returned type being correct with this
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/exceptions.py
```diff
@@ -14,9 +14,16 @@
...
 class BluepySnapError(Exception):
     """Base SNAP exception."""
+
+
+class BluepySnapDeprecationError(Exception):
+    """SNAP deprecation exception."""
+
+
+class BluepySnapDeprecationWarning(UserWarning):
+    """SNAP Deprecation warning."""
```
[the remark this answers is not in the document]
tomdele: done

bluepysnap/network.py
tomdele: no need when derived from abc. See : https://docs.python.org/3/library/abc.html#abc.abstractmethod
```
A class that has a metaclass derived from ABCMeta cannot be instantiated unless all of its abstract methods and properties are overridden.
```

bluepysnap/network.py
tomdele: done

bluepysnap/network.py
tomdele: indeed

bluepysnap/nodes.py
tomdele: This is something I would like to change when the properties/get from the edges remove the possibility to return the edge_ids. 
Once this is done, I will change the signatures accordingly.

bluepysnap/network.py
tomdele: The rational being this function, compared to use a class object or something similar, was the single use of the Storage in the whole class inside _get_populations (which is not public). Plus I will need to define the config and so to override the _get_population anyway or to add a Class object / Class member 
So I think it is more convoluted this way and I don t see the possible usage of a : 
Nodes.Storage or Edges.Storage for the user.

bluepysnap/network.py
wizmer: Ah yes. My bad !

bluepysnap/network.py
wizmer: I'm good with this. Thanks

wizmer (approved):
Very nice.

tomdele:
I hope this is one of the last ultra massive review for snap. 
Thanks a lot again for your feedback and help on the different difficulties included in this review.

asanin-epfl:
Thank you for such enormous work!

PR #119, opened 2021-01-25
Fix "Returning ids with get/properties" warning

tomdele:
- Fix warning when using edges functions using pathway_edges as backend.
- Use DeprecationWarning as base class for BluepySnapDeprecationWarning


asanin-epfl (approved):

tomdele:
Thank you !

PR #121, opened 2021-01-27
Ensure the dtype for the ids

tomdele:
It is quite unsafe to the use uint64 for the ids due to the numpy issue:
https://github.com/numpy/numpy/issues/15084

Where :
numpy.uint64 + int --> float64
numpy.uint64 += int --> float64

So it fixes the ids dtype to int64 until the issue in numpy is fixed.

- add a IDS_DTYPE to contain the ids dtype
- add a ensure_ids function
- add ensure_ids when ids are returned

### DIFF
--- a/bluepysnap/utils.py
+++ b/bluepysnap/utils.py
@@ -29,7 +29,8 @@ from bluepysnap.circuit_ids import CircuitNodeId, CircuitEdgeId
 from bluepysnap.sonata_constants import DYNAMICS_PREFIX
 
 # dtypes for the different node and edge ids. We are using np.int64 to avoid the infamous
-# https://github.com/numpy/numpy/issues/15084 numpy problem.
+# https://github.com/numpy/numpy/issues/15084 numpy problem. This type needs to be used for
+# all returned node or edge ids.
 IDS_DTYPE = np.int64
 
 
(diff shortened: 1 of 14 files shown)


tomdele:
(ping @eleftherioszisis for archngv)

asanin-epfl:
@tomdele please elaborate on the general policy of using this new util function `ensure_ids`. When it should be used? I still see some usages like `np.asarray(self._frame_population.get_node_ids(), dtype=np.int64)` in `frame_report.py` or `np.asarray(selection.flatten(), dtype=np.int64)` in `edges.py`.

tomdele:
I totally forgot the reports indeed. 
I tried to explain a bit better the policy. Let me know if this is clearer.

bluepysnap/utils.py
```diff
@@ -72,6 +73,8 @@ def ensure_ids(a):
     It is quite unsafe to the use uint64 for the ids due to this problem where :
     numpy.uint64 + int --> float64
     numpy.uint64 += int --> float64
+
+    This function needs to be used everywhere node_ids or edge_ids are returned.
```
asanin-epfl, line 77: Now this is very clear :D

asanin-epfl (approved):

asanin-epfl (commented):
Forgot one thing. Should be CHANGELOG.rst updated?

tomdele:
> Forgot one thing. Should be CHANGELOG.rst updated?

Hopefully you are here @asanin-epfl !

asanin-epfl (approved):

tomdele:
Thank you !

PR #122, opened 2021-02-09
Introduce NeuronModelsHelper, an analog of MorphHelper

asanin-epfl:
It is used to access neuron models.


asanin-epfl:
I don't get it. My 2nd commit is not represented here, wtf.

tomdele:
I think github is having troubles. 
The page css is fully broken for me.

PR #123, opened 2021-02-10
Introduce NeuronModelsHelper, an analog of MorphHelper

asanin-epfl:
It is used to access neuron models.


bluepysnap/neuron_models.py
```diff
@@ -0,0 +1,76 @@
...
+"""Neuron models access."""
+
+from pathlib import Path
+
+from bluepysnap.circuit_ids import CircuitNodeId
+from bluepysnap.exceptions import BluepySnapError
+from bluepysnap.sonata_constants import Node
+
+
+class NeuronModelsHelper(object):
```
tomdele, line 27: don t need the object

bluepysnap/morph.py
```diff
@@ -61,9 +61,9 @@ def get_filepath(self, node_id):
         Args:
             node_id (int/CircuitNodeId): could be a int or CircuitNodeId.
         """
-        name = self._population.get(node_id, Node.MORPHOLOGY)
         if isinstance(node_id, (int, CircuitNodeId)):
```
tomdele, line 64: We can : 
```
if not isinstance(node_id, (int, CircuitNodeId)):
     raise BluepySnapError("node_id must be a int or a CircuitNodeId")
name = self._population.get(node_id, Node.MORPHOLOGY)
return Path(self._morph_dir, f"{name}.swc")
```

bluepysnap/neuron_models.py
```diff
@@ -0,0 +1,76 @@
...
+    def get_filepath(self, node_id):
+        """Return path to model file corresponding to `node_id`.
+
+        Args:
+            node_id (int|CircuitNodeId): node id
+
+        Returns:
+            Path: path to the model file of neuron
+        """
+        if isinstance(node_id, (int, CircuitNodeId)):
```
tomdele, line 57: you can :
```
if not isinstance(node_id, (int, CircuitNodeId)):
    raise BluepySnapError("node_id must be a int or a CircuitNodeId")
```
to reduce the indentation level.

bluepysnap/neuron_models.py
```diff
@@ -0,0 +1,76 @@
...
+            node = self._population.get(node_id, [Node.MODEL_TYPE, Node.MODEL_TEMPLATE])
+            if node[Node.MODEL_TYPE] == "biophysical":
+                models_dir = self._config_components.get("biophysical_neuron_models_dir")
```
tomdele, line 60: This is so unclear for me from the specs if `"biophysical_neuron_models_dir"` is an expected key or not ...

bluepysnap/neuron_models.py
```diff
@@ -0,0 +1,76 @@
...
+                if models_dir is None:
+                    raise BluepySnapError(
+                        "Missing 'biophysical_neuron_models_dir' in Sonata config")
+            else:
+                models_dir = self._config_components.get("point_neuron_models_dir")
```
tomdele, line 65: Same for `"point_neuron_models_dir"`

bluepysnap/neuron_models.py
```diff
@@ -0,0 +1,76 @@
...
+                if models_dir is None:
+                    raise BluepySnapError("Missing 'point_neuron_models_dir' in Sonata config")
+
+            template = node[Node.MODEL_TEMPLATE]
+            assert ':' in template, "Format of 'model_template' must be <schema>:<resource>."
+            schema, resource = template.split(':', 1)
```
tomdele, line 71: Shouldn't we test if there are more than one `:` (like when someone is using colons in the resource name)

tests/test_neuron_models.py
```diff
@@ -0,0 +1,92 @@
...
+from bluepysnap.sonata_constants import Node
+from bluepysnap.exceptions import BluepySnapError
+
+import pytest
+
+from utils import TEST_DATA_DIR, copy_circuit, copy_config, edit_config, create_node_population
+
+
+def test_invalid_model_type():
+    with copy_circuit() as (circuit_copy_path, config_copy_path):
```
tomdele, line 14: Can you add a comment with the reason of the fail ?

tests/test_neuron_models.py
```diff
@@ -0,0 +1,92 @@
...
+        test_obj.get_filepath('1')
+    assert "node_id must be a int or a CircuitNodeId" in e.value.args[0]
+
+
+def test_get_filepath_biophysical():
+    nodes = create_node_population(str(TEST_DATA_DIR / 'nodes.h5'), "default")
+    config = Config(TEST_DATA_DIR / 'circuit_config.json').resolve()
+    test_obj = test_module.NeuronModelsHelper(config['components'], nodes)
+
+    node_id = 0
```
tomdele, line 45: Can you test with a CircuitId ? 
```
node_id = CircuitNodeId("default", 0)
```

tomdele: And maybe with 
```
node_id = CircuitNodeId("default2", 0)
```
Just to check what is returned (I most probably forgot to test this in the MorphHelper also)

bluepysnap/neuron_models.py
asanin-epfl: It is for 'biophysical' nodes

bluepysnap/neuron_models.py
asanin-epfl: It is required for 'point_neuron' neurons

bluepysnap/neuron_models.py
asanin-epfl: `split` happens by the first ':'. What is after is a different problem. It's up to a user to check it. The file can be missing even. There is also a URL possible but I didn't validate for it. URL seems speculative to me at the moment.

tests/test_neuron_models.py
asanin-epfl: Nope, you didn't forget to test. I didn't test it because this check seemed to me outside of this context. Should not be enough of tests that check different usages of `CircuitNodeId` separately?

bluepysnap/neuron_models.py
tomdele: thanks

bluepysnap/neuron_models.py
tomdele: Perfect makes sense

tests/test_neuron_models.py
tomdele: > Should not be enough of tests that check different usages of CircuitNodeId separately?

Not sure to understand what you mean here.

asanin-epfl: Then lets resolve as you requested.

tomdele (approved):

tomdele:
Looks good !

asanin-epfl:
Wonderful, thank you!

PR #127, opened 2021-02-12
Move query functionality to a standalone module

asanin-epfl:
So it can be reused by edges.
Currently it's an incomplete draft. @tomdele what are your thoughts?

### DIFF
--- a/bluepysnap/nodes.py
+++ b/bluepysnap/nodes.py
@@ -30,13 +30,10 @@ from cached_property import cached_property
 from bluepysnap.network import NetworkObject
 from bluepysnap import utils
 from bluepysnap.exceptions import BluepySnapError
-from bluepysnap.sonata_constants import (DYNAMICS_PREFIX, NODE_ID_KEY,
-                                         POPULATION_KEY, Node, ConstContainer)
+import bluepysnap.query as query
+from bluepysnap.sonata_constants import (DYNAMICS_PREFIX, Node, ConstContainer)
 from bluepysnap.circuit_ids import CircuitNodeId, CircuitNodeIds
 from bluepysnap._doctools import AbstractDocSubstitutionMeta
-
-# this constant is not part of the sonata standard
-NODE_SET_KEY = "$node_set"
 
 
 class Nodes(NetworkObject, metaclass=AbstractDocSubstitutionMeta,
@@ -217,18 +214,6 @@ class NodeStorage:
         return result
 
 
-# TODO: move to `libsonata` library
-def _complex_query(prop, query):
-    # pylint: disable=assignment-from-no-return
-    result = np.full(len(prop), True)
-    for key, value in query.items():
-        if key == '$regex':
-            result = np.logical_and(result, prop.str.match(value + "\\Z"))
-        else:
-            raise BluepySnapError("Unknown query modifier: '%s'" % key)
-    return result
-
-
 class NodePopulation:
     """Node population access."""
 
@@ -385,89 +370,17 @@ class NodePopulation:
             raise BluepySnapError("Undefined node set: '%s'" % node_set_name)
         return self._node_sets[node_set_name]
 
-    def _positional_mask(self, node_ids):
-        """Positional mask for the node IDs.
+    def _resolve_nodesets(self, queries):
+        def _resolve(queries, queries_key):
+            if queries_key == query.NODE_SET_KEY:
+                if query.AND_KEY not in queries:
+                    queries[query.AND_KEY] = []
+                queries[query.AND_KEY].append(self._get_node_set(queries[queries_key]))
+                del queries[queries_key]
 
-        Args:
-            node_ids (None/numpy.ndarray): the ids array. If None all ids are selected.
-
-        Examples:
-            if the data set contains 5 nodes:
-            _positional_mask([0,2]) --> [True, False, True, False, False]
-        """
-        if node_ids is None:
-            return np.full(len(self._data), fill_value=True)
-        mask = np.full(len(self._data), fill_value=False)
-        valid_node_ids = pd.Index(utils.ensure_list(node_ids)).intersection(self._data.index)
-        mask[valid_node_ids] = True
-        return mask
-
-    def _circuit_mask(self, queries):
-        """Handle the population, node ID and node set queries."""
-        populations = queries.pop(POPULATION_KEY, None)
-        if populations is not None and self.name not in set(utils.ensure_list(populations)):
-            node_ids = []
-        else:
-            node_ids = queries.pop(NODE_ID_KEY, None)
-        node_set = queries.pop(NODE_SET_KEY, None)
-        if node_set is not None:
-            if not isinstance(node_set, str):
-                raise BluepySnapError("{} is not a valid node set name.".format(node_set))
-            node_ids = node_ids if node_ids else self._data.index.values
-            node_ids = np.intersect1d(node_ids, self.ids(node_set))
-        return queries, self._positional_mask(node_ids)
-
-    def _properties_mask(self, queries, raise_missing_prop):
-        """Return mask of node IDs with rows matching `props` dict."""
-        # pylint: disable=assignment-from-no-return
-        circuit_keys = {POPULATION_KEY, NODE_ID_KEY, NODE_SET_KEY}
-        unknown_props = set(queries) - set(self._data.columns) - circuit_keys
-        if unknown_props:
-            if raise_missing_prop:
-                raise BluepySnapError(
-                    "Unknown node properties: [{0}]".format(", ".join(unknown_props)))
-            return np.full(len(self._data), fill_value=False)
-
-        queries, mask = self._circuit_mask(queries)
-        if not mask.any():
-            # Avoid fail and/or processing time if wrong population or no nodes
-            return mask
-
-        for prop, values in queries.items():
-            prop = self._data[prop]
-            if np.issubdtype(prop.dtype.type, np.floating):
-                v1, v2 = values
-                prop_mask = np.logical_and(prop >= v1, prop <= v2)
-            elif isinstance(values, str) and values.startswith('regex:'):
-                prop_mask = _complex_query(prop, {'$regex': values[6:]})
-            elif isinstance(values, Mapping):
-                prop_mask = _complex_query(prop, values)
-            else:
-                prop_mask = np.in1d(prop, values)
-            mask = np.logical_and(mask, prop_mask)
-        return mask
-
-    def _operator_mask(self, queries, raise_missing_prop):
-        """Handle the query operators '$or', '$and'."""
-        if len(queries) == 0:
-            return np.full(len(self._data), True)
-
-        # will pop the population and or/and operators so need to copy
-        queries = deepcopy(queries)
-        first_key = list(queries)[0]
-        if first_key == '$or':
-            queries = queries.pop("$or")
-            operator = np.logical_or
-        elif first_key == '$and':
-            queries = queries.pop("$and")
-            operator = np.logical_and
-        else:
-            return self._properties_mask(queries, raise_missing_prop)
-
-        mask = np.full(len(self._data), first_key != "$or")
-        for query in queries:
-            mask = operator(mask, self._operator_mask(query, raise_missing_prop))
-        return mask
+        resolved_queries = deepcopy(queries)
+        query.traverse_queries_bottom_up(resolved_queries, _resolve)
+        return resolved_queries
 
     def _node_ids_by_filter(self, queries, raise_missing_prop):
         """Return node IDs if their properties match the `queries` dict.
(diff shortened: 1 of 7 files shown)


bluepysnap/query.py
```diff
@@ -0,0 +1,116 @@
...
+            prop_mask = np.logical_and(prop >= v1, prop <= v2)
+        elif isinstance(values, str) and values.startswith('regex:'):
+            prop_mask = np.logical_and(np.full(len(prop), True), prop.str.match(values[6:] + "\\Z"))
+        else:
+            prop_mask = np.in1d(prop, values)
+        mask = np.logical_and(mask, prop_mask)
+    return mask
+
+
+def traverse_queries_bottom_up(queries, node_function):
```
asanin-epfl, line 72: The key change is this traversing function

bluepysnap/nodes.py
```diff
@@ -217,18 +214,6 @@ def load_population_data(self, population):
         return result
 
 
-# TODO: move to `libsonata` library
-def _complex_query(prop, query):
-    # pylint: disable=assignment-from-no-return
-    result = np.full(len(prop), True)
-    for key, value in query.items():
-        if key == '$regex':
```
asanin-epfl, line 225: $regex node have been dropped, only `regex:` prefix is kept.

tomdele: No this should be the opposite. `regex:` should be dropped. This is a bad idea to have the function in the argument. See basically my comment here : 
https://github.com/BlueBrain/libsonata/issues/119#issuecomment-769052911

tomdele:
I think this is definitely a great idea. Having the query system in a module makes sense and specially if we want to use it with the edges also.

bluepysnap/query.py
tomdele: I think it clearer than the current implementation

bluepysnap/nodes.py
asanin-epfl: I don't see any definition of `$regex` in https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#node-sets-file. I agree with your point. How can we be sure that this would be the only way to interpret? For example what about `{'$regex': {Cell.MTYPE: '.*BP'} }`? This way we can have multiple of regex fields like `{'$regex': {Cell.MTYPE: '.*BP', Cell.ETYPE: '.*bap'} }`. What are your thoughts?

tomdele: The queries are not part of the sonata specs at all. This is something we need to introduce as part of the bluepysnap extra features.
The scope of the `$or` / `$and` differs from the normal scope of the queries. They are "queries operators" as they act on queries themselves and not on a field. Like a "$not" operator can do also.
All the other operators are defined the same way : {the field : {the operator: the parameters}}. With some implicit operators for the most common ones like the equal : 
{"property": "the value to match"}

So to me : 
`{'$regex': {Cell.MTYPE: '.*BP', Cell.ETYPE: '.*bap'} }` should be avoided and we should use:
```
 {$and: [{Cell.MTYPE: `{'$regex':  '.*BP}}, {Cell.ETYPE: {'$regex':'.*bap'}}]
```
With this everything is well defined. And you can easily combine the queries in a homogeneous way : 
```
 {$and: [{Cell.MTYPE: `{'$regex':  '.*BP}}, {Cell.ETYPE: {'$regex':'.*bap'}}, {Node.LAYER: [2, 3]}, {Node.X: {"$lt": 2}}]}
```
Also, you can easily do something like : 
```
query_1 = {Cell.MTYPE: `{'$regex':  '.*BP}}
query_2 = {Cell.ETYPE: {'$regex':'.*bap'}
query_3 = {Node.LAYER: [2, 3]}
query_4 = {Node.X: {"$lt": 2}}
full = {$and: [query_1, query_2,query_3, query_4]}
```
and you have only one way of doing it no matter the operator used for each parts.

What I mean is, if you have :
```
query_1 = {Cell.MTYPE: `{'$regex':  '.*BP}}
query_2 = {Cell.ETYPE: {'$regex':'.*bap'}
```
or 
```
query_1 = {Cell.MTYPE: `{'$regex':  '.*BP}}
query_2 = {Cell.LAYER: {'$gt':4}
```
the combination will always be : 
```
full = {$and: [query_1, query_2]}
```

bluepysnap/query.py
```diff
@@ -0,0 +1,159 @@
...
+    """Traverse queries tree from leaves to root, left to right.
+
+    Args:
+        queries (dict): queries
+        node_function (function): function to execute on each node of queries tree in traverse order
+    """
+    for key in queries.copy():
```
asanin-epfl, line 94: this `.copy` is a bit tricky. It is needed only because `queries` are manipulated in `_resolve_nodesets`.

asanin-epfl:
I just wanted to make CI green. Now I will move part of tests to `test_query`

bluepysnap/query.py
tomdele: Yes I had the same problem in the old impl

tomdele:
Ok let me know when this is ready to review. At the first glance this is solid already

asanin-epfl:
it's ready, please review

bluepysnap/query.py
```diff
@@ -0,0 +1,159 @@
...
+    return mask
+
+
+def _circuit_mask(data, population_name, queries):
+    """Handle the population, node ID queries."""
+    populations = queries.pop(POPULATION_KEY, None)
+    if populations is not None and population_name not in set(utils.ensure_list(populations)):
+        ids = []
+    else:
+        ids = queries.pop(NODE_ID_KEY, queries.pop(EDGE_ID_KEY, None))
```
asanin-epfl, line 57: I want to accent here as `node_id` and `edge_id` become a general id and can be used interchangeably.

bluepysnap/nodes.py
```diff
@@ -385,89 +370,17 @@ def _get_node_set(self, node_set_name):
...
+                    queries[query.AND_KEY] = []
+                queries[query.AND_KEY].append(self._get_node_set(queries[queries_key]))
+                del queries[queries_key]
 
-        mask = np.full(len(self._data), first_key != "$or")
-        for query in queries:
-            mask = operator(mask, self._operator_mask(query, raise_missing_prop))
-        return mask
+        resolved_queries = deepcopy(queries)
+        query.traverse_queries_bottom_up(resolved_queries, _resolve)
```
tomdele, line 382: The more I read your code the more I think having a Query class would make sense. 
What do you think ?

bluepysnap/query.py
```diff
@@ -0,0 +1,159 @@
...
+
+def _positional_mask(data, ids):
+    """Positional mask for the node IDs.
+
+    Args:
+        ids (None/numpy.ndarray): the ids array. If None all ids are selected.
+
+    Examples:
+        if the data set contains 5 nodes:
+        _positional_mask([0,2]) --> [True, False, True, False, False]
```
tomdele, line 41: I made a bad doc here: 
```
_positional_mask(data, [0,2]) --> [True, False, True, False, False]
```

bluepysnap/query.py
```diff
@@ -0,0 +1,159 @@
...
+        # Avoid fail and/or processing time if wrong population or no nodes
+        return mask
+
+    for prop, values in queries.items():
+        prop = data[prop]
+        if np.issubdtype(prop.dtype.type, np.floating):
+            v1, v2 = values
+            prop_mask = np.logical_and(prop >= v1, prop <= v2)
+        elif isinstance(values, str) and values.startswith('regex:'):
+            prop_mask = _complex_query(prop, {'$regex': values[6:]})
```
tomdele, line 78: I think we should remove this part. This is redundant with the Mapping part and I removed it from bleupy also.

bluepysnap/query.py
tomdele: I think the name is not perfect. 
What about : 
transform_queries_bottom_up(queries, transform_function)

Because the idea is to transform the queries using a transform_function from the bottom to up. 
Also this makes sense only if the current "node_function" is a inplace transform function of the queries if I understand correctly.

bluepysnap/query.py
```diff
@@ -0,0 +1,159 @@
...
+        mask = np.logical_and(mask, prop_mask)
+    return mask
+
+
+def traverse_queries_bottom_up(queries, node_function):
+    """Traverse queries tree from leaves to root, left to right.
+
+    Args:
+        queries (dict): queries
+        node_function (function): function to execute on each node of queries tree in traverse order
```
tomdele, line 92: We should have a generic name for the this parameter no ?

bluepysnap/query.py
```diff
@@ -0,0 +1,159 @@
...
+    def _collect(_, query_key):
+        if query_key not in ALL_KEYS:
+            props.add(query_key)
+
+    props = set()
+    traverse_queries_bottom_up(queries, _collect)
+    return props
+
+
+def operator_mask(data, population_name, queries):
```
tomdele, line 126: I think the name is not correct anymore (and maybe even before when it was in the NodePopulation class). Something like : 
resolve_ids or apply_query
could be better no ?

It would be more elegant also to have : 
def resolve_ids(population, query):
     data = population.get()
     population_name = population.name
or we can decide how to access a "public population.data". 

One problem I have right now is : how can we handle the data argument for the edges ? We will not be able to have a dataframe in memory. Do you want to use a dask dataframe ?

asanin-epfl: > I think the name is not correct anymore (and maybe even before when it was in the NodePopulation class).

`resolve_ids` sounds good. I kept old names so the review is easier.

> One problem I have right now is : how can we handle the data argument for the edges ? We will not be able to have a dataframe in memory. Do you want to use a dask dataframe ?

I planned to use a DataFrame that contains only properties in the query. It is easy to do with `query.get_properties` function.

bluepysnap/nodes.py
asanin-epfl: It is possible, I need to look into it. I kept old code as much as possible, so new changes are easier to see.

bluepysnap/query.py
asanin-epfl: Idea is not transform but to traverse. Transforming is collateral here due to resolving of `$node_set`.

bluepysnap/nodes.py
asanin-epfl: After some experiments. There is one problem. Such class must be used only within `.query` module. We will need to transform `queries` dict into an instance of this class. Such transformation would require `queries` traversal. In order to resolve ids, we would need to traverse the instance of this class. So, double traversal. I'm against double traversal. Possibly I picked the wrong class representation. Please suggest your alternatives:

```
class QueryTree:
    def __init__(self, query_dict, parent=None):
        self._parent = parent
        if parent is None:
            self._key = AND_KEY
            self._value = None
            self._children = []
            for key, value in query_dict.items():
                self._children.append(QueryTree({key: value}, self))
        else:
            self._key, self._value = unfold(query_dict)
            if self.key == OR_KEY:
                pass
            elif self.key == AND_KEY:
                pass
            elif regex_case:
                pass
```

tomdele: ah I did not think about that indeed ... 

I spent some times thinking about it and I don t have anything really clean in my mind. Also because the traverse function aims at transforming the query but without any expected content at the end (can be a dict with other queries or a dict with np arrays or whatever). 

Hum yes this is not that easy to have a good data structure... so let s keep it like this and we can make another path on this if we have a brilliant idea at some point.

tests/test_query.py
```diff
@@ -0,0 +1,54 @@
...
+                                  {"population": "default", "node_id": [2], "other": "val"})
+    assert queries == {"other": "val"}
+    npt.assert_array_equal(mask, [False, False, True])
+
+    queries, mask = _circuit_mask(data, "default", {"other": "val"})
+    assert queries == {"other": "val"}
+    npt.assert_array_equal(mask, [True, True, True])
+
+
+def test_operator_mask():
```
tomdele, line 35: name not correct anymore

tests/test_query.py
```diff
@@ -0,0 +1,54 @@
...
+    data = pd.DataFrame([
```
tomdele, line 36: I would recommend adding nested "$and" / "$or" queries just to check if you do the operator queries as expected.

tomdele: I did it in the node_population.ids() I think but it would be nice to have that here too

bluepysnap/query.py
asanin-epfl: Here is the function that does explicit transform + it does not require deep copying. From functions point of view it is more proper classic variant of the function. One problem with it. It is hard to read/understand:
```
def transform_queries_bottom_up(queries, transform):
    """Traverse queries tree from leaves to root, left to right.

    Args:
        queries (dict): queries
        transform (function): function to execute on each node of `queries` in traverse order
    """
    transformed_queries = {}
    for key in list(queries.keys()):
        if key in {OR_KEY, AND_KEY}:
            transformed_queries[key] = []
            for subquery in queries[key]:
                transformed_queries[key].append(transform_queries_bottom_up(subquery, transform))
            # `transform` has to be called on `transformed_queries`
            transformed_queries[key] = transform(transformed_queries, key)
        elif isinstance(queries[key], Mapping):
            if VALUE_KEYS & set(queries[key]):
                if not set(queries[key]).issubset(VALUE_KEYS):
                    raise BluepySnapError("Value operators can't be used with plain values")
            else:
                transformed_queries[key] = transform_queries_bottom_up(queries[key], transform)
                # `transform` has to be called on `transformed_queries`
                transformed_queries[key] = transform(transformed_queries, key)
        else:
            transformed_queries[key] = transform(queries, key)
    return transformed_queries
```

tests/test_query.py
asanin-epfl: I had one on the line 47. Added another one below it. Please verify.

tomdele: oh sorry I missed it indeed my bad

bluepysnap/query.py
tomdele: I see your point. 
Gosh, it appears to be more complex design wise than first expected (at least to me)!
Ok lets keep it like this. Maybe having the edges ids in the future will change the function again for some reasons. So I think your impl is good at the moment.

tomdele (approved):
I think this is legit.
Nice job, thanks a lot ! :+1:

asanin-epfl:
Thank you for feedback! I will merge it at noon.

PR #128, opened 2021-02-22
Improve perf for positional mask

tomdele:
- improve perfs for the positional mask
- positional mask still correct if ids are not monotonic starting from 0

Tests : 
```
df = pd.DataFrame(index=np.arange(5000000))

test_ids = np.random.randint(len(df), size=50)
%timeit _positional_mask_old(df, test_ids)
115 ms ± 910 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit _positional_mask(df, test_ids)
95.2 µs ± 417 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

test_ids = np.random.randint(len(df), size=50000)
%timeit _positional_mask_old(df, test_ids)
141 ms ± 3.89 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit _positional_mask(df, test_ids)
2.32 ms ± 87.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

test_ids = np.random.randint(len(df), size=3000000)
%timeit _positional_mask_old(df, test_ids)
1.05 s ± 15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit _positional_mask(df, test_ids)
111 ms ± 2.33 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

df = pd.DataFrame(index=np.arange(50000000))
test_ids = np.random.randint(len(df), size=3000000)
%timeit _positional_mask_old(df, test_ids)
2.45 s ± 24 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit _positional_mask(df, test_ids)
112 ms ± 2.25 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

PR #129, opened 2021-03-02
Implement queries mechanism for edges

asanin-epfl:


asanin-epfl:
Probable more tests are required. @tomdele Should we rather merge tests of queries from nodes and edges into one place?

bluepysnap/edges.py
```diff
@@ -480,7 +482,33 @@ def _get(self, selection, properties=None):
...
+            >>> self._edge_ids_by_filter({'$or': [{ Edge.PRE_X_CENTER: [2, 3]},
+            >>>                              { Edge.POST_SECTION_POS: (0, 1),
+            >>>                              Edge.SYN_WEIGHT: (0.,1.4) }]})
+
+        """
+        properties = query.get_properties(queries)
+        unknown_props = properties - self.property_names
+        if raise_missing_prop and unknown_props:
+            raise BluepySnapError(f"Unknown edge properties: {unknown_props}")
+        data = self.get(None, properties - unknown_props)
```
tomdele, line 507: I am quite scared this cannot be done. I allowed he None for the get because user is in full control but I am sure for huge circuit this will not be possible... Did you try ?

asanin-epfl: > I am quite scared this cannot be done.

Should I understand this as "I am quite scared that this can be done"?

> I allowed he None

Should I understand this as "I allowed the `None`"?

> Did you try ?

No but I agree that this would highly likely fail/stuck on edges of Tb size with many properties.

1. In general this should work when there are only several properties within the query. All group attributes in edges files are int or floats. Lets assume float64 => 8 bytes. Lets say we have 1e9 (billion) edges in one population. One property then would take 8Gb. I doubt that edges of that size will be processed on a desktop computer. So, we have a high performant server hardware. Even for 10 properties it will take 80Gb of Ram which is not much for server hardware.

2. I don't want to introduce Dask until we are sure that libsonata will not support it for 100%.

tomdele: I was not clear enough. 

What I meant is: 
Back in the days, the edges "queries" : `get`, `afferent_nodes`, `efferent_nodes` forced as entries a list of edge ids. Then I chose to allow a None list of edge ids which means : take all the edge_ids.
I did that because this is not hidden to the user and he clearly needs to call it:
Ex :
```
get(None)
```
And the None is never a default value.

Using the queries I am a bit scared this is a bit more hidden and unclear why one fills his memory. 

I totally agree with the hardware but maybe we can do it a bit differently without dask. 
What about doing the property queries by chunk ? 

Something like : 
```
res = []
ids = self.ids(None)
critical_size = 60000000  # can be something computed using the dtypes of the properties
for chunk in np.array_split(ids, len(ids) // critical_size):
    data = self.get(chunk, properties - unknown_props)
    res.extend(query.resolve_ids(data, self.name, queries))
```
? 
doing so you reduce a lot the memory consumption because you don t load everything in memory at the same time and I guess with the change I made in #128 you should not have problems with the positional queries.

asanin-epfl: Great idea with chunking, I like it. Let see how it behaves. The current approach failed for 1.6Tb files (9.12 billion edges) with option `--mem=100g` and a single property. Around 74Gb is needed for single property in this case. The error was expected: `Out Of Memory`.

tomdele: Thanks a lot for the check !
We will need to find a cleaver way of doing this so it can be usable ... I hope chunking can help.

asanin-epfl: `--mem=0` worked for this 1.6Tb case but it took `1109.78255582` seconds.

tomdele: yeah I did not expect insane perfs anyway.
Maybe we will be able to improve afterwards with multi processing.

bluepysnap/edges.py
```diff
@@ -480,7 +482,37 @@ def _get(self, selection, properties=None):
...
+        unknown_props = properties - self.property_names
+        if raise_missing_prop and unknown_props:
+            raise BluepySnapError(f"Unknown edge properties: {unknown_props}")
+        res = []
+        ids = self.ids(None)
+        chunk_size = int(1e8)
+        for chunk in np.array_split(ids, 1 + len(ids) // chunk_size):
+            data = self.get(chunk, properties - unknown_props)
+            res.extend(chunk[query.resolve_ids(data, self.name, queries)])
+        return res
```
tomdele, line 513: We should use numpy array with IDS_DTYPE instead. I wrote the snippet with a list but I forgot we return arrays ...

bluepysnap/edges.py
asanin-epfl: ok, chunking is working and helps. It worked for 1.6Tb with `--mem=200gb`. Ran actually faster for `799.776455279999` seconds

bluepysnap/edges.py
asanin-epfl: btw wdyt about
```
        res = []
        ids = self.ids(None)
        chunk_size = int(1e8)
        for chunk in np.array_split(ids, 1 + len(ids) // chunk_size):
            data = self.get(chunk, properties - unknown_props)
            res.append(chunk[query.resolve_ids(data, self.name, queries)])
        return np.concatenate(res, dtype=IDS_DTYPE)
```

tomdele: This needs to be benchmarked. I did this housand times but keep forgetting about it. What s the best between what you did and concatenate at every chunk :
```
        res = np.empty(dtype=IDS_DTYPE)
        ids = self.ids(None)
        chunk_size = int(1e8)
        for chunk in np.array_split(ids, 1 + len(ids) // chunk_size):
            data = self.get(chunk, properties - unknown_props)
            res.concatenate([res, chunk[query.resolve_ids(data, self.name, queries)]])
        return res
```
(or something equivalent I did not run this piece of code just writing it live) ?

tomdele: Did you manage to contain the memory consumption with the chunks?

asanin-epfl: > Did you manage to contain the memory consumption with the chunks?

Well, it requires less memory for sure. 1.6Tb edges can be processed with `--mem=200g` and chunked version. The non-chunked version can process only with `--mem=0` which is around 500g depending on the machine.

asanin-epfl: > This needs to be benchmarked.

Not much difference in terms of memory. Used the below benchmark
```
pip install memory_profiler
```

```python
from memory_profiler import profile
import numpy as np


np.random.seed(666)

@profile
def benchmark_append():
    res = []
    for i in range(10):
        start = np.random.randint(-10000, 10000)
        end = np.random.randint(100000, 1000000)
        res.append([i for i in range(start, end)])
    return np.concatenate(res)

@profile
def benchmark_extend():
    res = []
    for i in range(10):
        start = np.random.randint(-10000, 10000)
        end = np.random.randint(100000, 1000000)
        res.extend([i for i in range(start, end)])
    return np.array(res)

if __name__ == '__main__':
    benchmark_append()
    benchmark_extend()
```

bluepysnap/edges.py
```diff
@@ -480,7 +482,37 @@ def _get(self, selection, properties=None):
...
+        unknown_props = properties - self.property_names
+        if raise_missing_prop and unknown_props:
+            raise BluepySnapError(f"Unknown edge properties: {unknown_props}")
+        res = []
+        ids = self.ids(None)
+        chunk_size = int(1e8)
+        for chunk in np.array_split(ids, 1 + len(ids) // chunk_size):
+            data = self.get(chunk, properties - unknown_props)
+            res.extend(chunk[query.resolve_ids(data, self.name, queries)])
+        return np.array(res, dtype=IDS_DTYPE)
```
tomdele, line 513: do you think we could have better perfs using smaller but more chunks and using multiprocessing (not dask just multiprocess)?

asanin-epfl: I don't know. The problem here is the layer of libsonata between us and HDF5. If we go directly via `h5py` then such optimization would work. Otherwise, I am not sure.

tomdele: Hum good points ... Ok we can keep it like this.

bluepysnap/edges.py
```diff
@@ -480,7 +482,37 @@ def _get(self, selection, properties=None):
...
+            >>>                              Edge.SYN_WEIGHT: (0.,1.4) }]})
+
+        """
+        properties = query.get_properties(queries)
+        unknown_props = properties - self.property_names
+        if raise_missing_prop and unknown_props:
+            raise BluepySnapError(f"Unknown edge properties: {unknown_props}")
+        res = []
+        ids = self.ids(None)
+        chunk_size = int(1e8)
```
tomdele, line 509: We can keep it like this for the moment but do you think we can keep it close to ~10Gb of memory usage in the future ?

bluepysnap/edges.py
asanin-epfl: It is not necessary to merge it. We can abandon. I've implemented rather for history so we can have this discussion. If you still have some doubts, lets postpone or even abandon.

bluepysnap/edges.py
asanin-epfl: Do you mean to detect the amount of available RAM by cleaver usage?

bluepysnap/edges.py
tomdele: No I think this is good to have this feature and we can improve it once it is here already. 

My only concern is to crash people pc when they are doing something they don t intend to. So the best would be to have something that can evaluate the size of the returned property dataframe in the function and compare it to the size of the user memory. So we can throw and clean memory before the ram + swap are full.

bluepysnap/edges.py
tomdele: yes I was writing it here : 
https://github.com/BlueBrain/snap/pull/129#discussion_r586322548

What do you think ?

tomdele: I don t have a super strong opinion. I just hate when my pc freeze when it was not clearly expected.

asanin-epfl: Please confirm
- detect available RAM memory
- detect size of edges file
- raise an exception if RAM is too low
- otherwise calculate chunk size, and start processing

tomdele: I would say yes with these functions : get_available_ram(), get_dataframe_memory() in utils.py maybe.

But what do you think about it ? Because this is doable but do you think this is "elegant" or even useful ? or should we just don t care about that ?

asanin-epfl: In general it is useful. However it is ahead of time in my opinion.

asanin-epfl: RAM checks

tomdele: Ok fine with me

tomdele (approved):

bluepysnap/edges.py
asanin-epfl: I'm still going to implement them here :)

asanin-epfl: ok, I've checked with this code

```python
        from psutil import virtual_memory, swap_memory

        properties = properties - unknown_props
        ids = self.ids(None)
        available_memory = swap_memory().free + virtual_memory().available
        max_required_memory = len(properties) * 8 * len(ids)  # assume 8 bytes per edge value
        if available_memory < max_required_memory:
            raise BluepySnapError('Not enough memory for edges querying')
        print(f'available_memory {available_memory}, max_required_memory {max_required_memory}')
        chunk_size = int(1e9)
        for chunk in np.array_split(ids, 1 + len(ids) // chunk_size):
            data = self.get(chunk, properties)
            res.extend(chunk[query.resolve_ids(data, self.name, queries)])
        return np.array(res, dtype=IDS_DTYPE)
```

It does not work because `libsonata` fails before we even reach this. So if there is not enough memory then `libsonata` will fail at
```
    @cached_property
    def _population(self):
        return self._edge_storage.storage.open_population(self.name)
```
Not much point in this check because if there is not enough memory then we don't reach it.

tomdele: ok really good to know.

tomdele: Thanks for the check. 
I must say I am not sure what we can do to avoid filling memory at the moment. So let s merge it like this and we will see. 
Maybe reducing the chunk size to 1e7 ?

asanin-epfl: ok but I want to emphasize again that if there is not enough memory => we fail at `libsonata` usage before reaching this code. Basically when we in `def _edge_ids_by_filter`, we already have enough memory.

tomdele: ahhh ok. Sorry that all good then !

tomdele: :+1:

asanin-epfl: Thank you so much. It was very helpful and useful.

tomdele:
Thanks a lot for this is really nice !

PR #133, opened 2021-04-20
Moving to neuromV2

tomdele:


asanin-epfl (approved):
so, do we have neurom as dependency only because of tests?

tomdele:
No we have the reader also in the morph helper and the get which returns a neurom morphology

tomdele:
the api did not change so I did not need to change the piece of code in the morphHelper. And I think this `get` function is useful due to the transformation matrix that is not 100% trivial we apply on the morphology.

asanin-epfl:
ah, ok. Thank you for clarifying!

PR #134, opened 2021-04-22
Use morphio for the morphologies

tomdele:
* removing the MORPH_CACHE_SIZE
* removing neurom as the main reader for morphologies
* adding morphio as the main reader for the morphologies


bluepysnap/morph.py
```diff
@@ -20,9 +20,10 @@
 from pathlib import Path
 
 import numpy as np
-import neurom as nm
+from morphio.mut import Morphology
```
asanin-epfl, line 23: Should we rather return read-only morphology? If user wants, he can call `as_mutable` on it.

bluepysnap/morph.py
```diff
@@ -47,10 +48,7 @@ def __init__(self, morph_dir, population):
         # all nodes from a population must have the same model type
         if not self._is_biophysical(0):
             raise BluepySnapError("Node population does not contain biophysical nodes.")
-        self._load = nm.load_neuron
-        if MORPH_CACHE_SIZE is not None:
-            from functools import lru_cache
-            self._load = lru_cache(maxsize=MORPH_CACHE_SIZE)(self._load)
+        self._load = Morphology
```
asanin-epfl, line 51: I'm looking at this strange function assigning, and don't see benefits. Should we just directly call `Morphology` in `get` function?

asanin-epfl (approved):
Looks very good. My comments are only for discussion.

bluepysnap/morph.py
asanin-epfl: don't answer, I see now. it's needed for transformation.

tomdele: I need to use a mut as a loader because I will apply transformations on it. 
I am then calling `as_immutable` on it at the end of the get: 
https://github.com/BlueBrain/snap/pull/134/files#diff-93c045373ea66b3e2b3de4f54a4610d3c1b80afbc993da10756c4186ac2aa3c8R82

bluepysnap/morph.py
tomdele: I think you are right. I kept it because I was like : 
"ok this a bit better if we want to change the loader again latter on". 
But yes I don t find a lot of good usage for it. 

I will remove. Thanks!

asanin-epfl (approved):

tomdele:
Thanks a lot !

bluepysnap/morph.py
tomdele: Oups I responded before seeing your comment

PR #137, opened 2021-05-25
Fix the morphology/model access using a numpy int

tomdele:
* Fix the morphology access using a numpy int
* Add tests for the int accesses

### DIFF
--- a/bluepysnap/utils.py
+++ b/bluepysnap/utils.py
@@ -53,6 +53,15 @@ def load_json(filepath):
         return json.load(f)
 
 
+def is_node_id(node_id):
+    """Check if node_id can be a valid node id.
+
+    Returns:
+        bool: true is node_id is a int, np.integer or CircuitNodeId. False for anything else.
+    """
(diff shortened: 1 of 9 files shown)


tomdele:
@eleftherioszisis you can have a look if needed

bluepysnap/utils.py
```diff
@@ -53,6 +53,15 @@ def load_json(filepath):
         return json.load(f)
 
 
+def is_node_id(node_id):
+    """Check if node_id can be a valid node id.
+
+    Returns:
+        bool: true is node_id is a int, np.integer or CircuitNodeId. False for anything else.
+    """
+    return isinstance(node_id, (int, np.integer, CircuitNodeId))
```
asanin-epfl, line 62: Just thinking here. If there is a dedicated method then should it also check for negative numbers? I guess should not.

Also can this function be used in https://github.com/BlueBrain/snap/blob/master/bluepysnap/nodes.py#L626 ?

asanin-epfl (approved):

tomdele:
Thank you !

PR #142, opened 2021-07-28
Added functionalities regarding the newest addition to BBP sonata spec.

joni-herttuainen:
* alternate morphology directories
* population config overwriting "components"
* population type available in NodePopulation/EdgePopulation
* added different edge and node types to bluepysnap/bbp.py
* updated circuit_validator to be closer to BBP spec


bluepysnap/edges.py
```diff
@@ -390,6 +400,16 @@ def _nodes(self, population_name):
...
 
+    @property
+    def config(self):
+        """Population config dictionary."""
+        return self._config
+
+    @property
+    def type(self):
+        """Population type."""
+        return self.config.get('type', DEFAULT_EDGE_TYPE)
```
joni-herttuainen, line 411: My question was whether we should use the `DEFAULT_EDGE_TYPE` here if `type` is not given or get it from the h5 files but I just realized that edges don't have `model_type` in them so the only way to get it right would be to deduce it by using the class definitions in `bbp.py`

bluepysnap/config.py
```diff
@@ -23,6 +23,19 @@
 from bluepysnap import utils
 from bluepysnap.exceptions import BluepySnapError
 
+EXPECTED_PATH_KEYS = ['morphologies_dir',
```
asanin-epfl, line 26: Why hardcoding this? Do we always expect relative paths for them? What if an absolute path comes under these keys?

bluepysnap/edges.py
asanin-epfl: I think the current implementation is fine. Class definitions from `bbp.py` can be used if you want to validate the value of `self.config['type']`. Do we want to validate it? I am not sure.

asanin-epfl (commented):
That is a general overview

bluepysnap/bbp.py
```diff
@@ -17,7 +17,217 @@
...
 """BBP cell / synapse attribute namespace."""
 
-from bluepysnap.sonata_constants import DYNAMICS_PREFIX, Edge, Node
+from bluepysnap.sonata_constants import DYNAMICS_PREFIX, ConstContainer, Edge, Node
+
+NODE_TYPES = ['biophysical', 'virtual', 'astrocyte', 'single_compartment', 'point_neuron']
+EDGE_TYPES = ['chemical', 'electrical', 'synapse_astrocyte', 'endfoot']
+
+
+class Astrocyte(ConstContainer):
```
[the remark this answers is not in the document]
joni-herttuainen: They're found in the [Circuit Documentation](https://bbpteam.epfl.ch/documentation/projects/Circuit%20Documentation/latest/sonata_tech.html), and these basically represent the ones listed as mandatory there. These are actually currently not used but my idea was to check the files for these fields in `circuit_validation` and then throw an error if any of them is missing. 

I would also expect them to inherit the `Node` (or `Edge` in case of edges) but these don't all have the same attributes defined in `sonata_constants`, so I am not sure what would be the best practice here.

bluepysnap/bbp.py
```diff
@@ -17,7 +17,217 @@
...
+
+    MORPHOLOGY = "morphology"  #:
+    MORPH_CLASS = "morph_class"
+    MODEL_TEMPLATE = "model_template"  #:
+    MODEL_TYPE = "model_type"  #:
+
+    MTYPE = "mtype"  #:
+
+    MODEL_TYPE = "model_type"  #:
+    MODEL_TEMPLATE = "model_template"  #:
```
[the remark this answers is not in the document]
joni-herttuainen: True, odd that no linter picked that.

bluepysnap/config.py
joni-herttuainen: No, this is to momentarily solve the issue of how to handle relative filenames that are not defined with leading `./`. Momentarily meaning whenever libsonata has config "parser" for simulations (they already have for circuits) I will deprecate the whole `bluepysnap.config` submodule.

asanin-epfl: That is great to hear. If `libsonata` can be reused then it is great. Ok to have it as temporary.

bluepysnap/edges.py
joni-herttuainen: Yeah, on the other hand, our main point is to provide API to access the data and it is users' responsibility to ensure the data is in the expected format. On the other hand though, we kind of maintain the format and it would be nice if users could easily verify that their files and config is what is expected. 

Should it be included in the API? My intuition is no but I also see the logic of shipping them together: Changes in the validation and API would go hand in hand.

asanin-epfl: The current codestyle of the project that such validation is located within `validation` module. Lets keep consistent codestyle.

eleftherioszisis: Given that development in circuit building and the addition of new features to the sonata configs could easily lead to inconsistent or deprecated configs, imho validation is crucial to remain in snap.

bluepysnap/morph.py
```diff
@@ -52,26 +58,46 @@ def __init__(self, morph_dir, population):
...
+        if not alternate_key:
+            raise BluepySnapError(f"Unsupported extension: {extension}")
+
+        morph_dir = self._alternate_morph_dir.get(alternate_key)
+        if not morph_dir:
+            raise BluepySnapError(f"'{alternate_key}' is not defined in 'alternate_morphologies'")
+
+        return morph_dir
+
+    def get_filepath(self, node_id, extension='swc'):
```
joni-herttuainen, line 78: I don't like this `extension` kwarg at all but having different functions for the alternate morphs did not make sense either. ArchNGV needs this to return a path, though. So, if anyone happens to come up with a better solution, please let me know and I'll happily change this one.

asanin-epfl:
I've pushed some new changes, @joni-herttuainen FYI.

joni-herttuainen:
Cool, thanks!

asanin-epfl:
What version should be used for those changes? 1.0.0 or 0.13.0?

joni-herttuainen:
Good question, I was trying to keep this as a minor release w/ the backwards compatibility. So, I think rather 0.13.0 than 1.0.0?

asanin-epfl:
I'm fine with 0.13.0

bluepysnap/bbp.py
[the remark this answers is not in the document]
asanin-epfl: I'm hesitating here. I see so much code duplication among those enums. @mgeplf what are your thoughts here?

bluepysnap/circuit_validation.py
```diff
@@ -143,6 +145,8 @@ def _check_required_datasets(config):
         types_file = nodes_dict.get('node_types_file')
         if types_file is not None and not Path(types_file).is_file():
             errors.append(fatal('Invalid "node_types_file": {}'.format(types_file)))
+        if not nodes_dict.get('populations'):
```
asanin-epfl, line 148: I thought it is optional, even in the new spec.

bluepysnap/circuit_validation.py
```diff
@@ -308,48 +313,81 @@ def _check_rotations():
     if missing_fields:
         errors.append(fatal('Group {} of {} misses biophysical fields: {}'.
                             format(group_name, group.file.filename, missing_fields)))
-    if 'components' not in config:
-        errors.append(fatal('No "components" in config'))
-        return errors
-    components = config['components']
-    errors += _check_components_dir('morphologies_dir', components)
-    errors += _check_components_dir('biophysical_neuron_models_dir', components)
+    _check_rotations()
```
asanin-epfl, line 316: Why bring it back here? I moved it on purpose in the last commit to `master` branch.

bluepysnap/circuit_validation.py
```diff
@@ -308,48 +313,81 @@ def _check_rotations():
...
     elif group_df['model_type'][0] == 'biophysical':
-        return _check_bio_nodes_group(group_df, group, config)
+        if 'components' not in config:
+            return errors + [fatal('No "components" in config')]
+        population = {**config['components'], **population}
+        return errors + _check_bio_nodes_group(group_df, group, population)
+    return errors
+
+
+def _check_populations_dict(populations_dict, populations_h5, file_name):
```
asanin-epfl, line 376: Can we please use `populations_config` instead of `populations_dict`?

bluepysnap/circuit_validation.py
```diff
@@ -371,6 +409,10 @@ def _check_nodes_population(nodes_dict, config):
         nodes = _get_h5_data(h5f, 'nodes')
         if not nodes or len(nodes) == 0:
             return [fatal('No "nodes" in {}.'.format(nodes_file))]
+        populations_dict = nodes_dict.get('populations', dict())
```
asanin-epfl, line 412: I'm surprised that lint didn't notice a plain `dict()`. In all recent PRs it requires it to be a literal.

asanin-epfl:
We are waiting for @mgeplf or @GianlucaFicarelli

bluepysnap/morph.py
```diff
@@ -42,36 +47,50 @@ def __init__(self, morph_dir, population):
         Returns:
             MorphHelper: A MorphHelper object.
         """
-        self._morph_dir = morph_dir
+        self._morph_dir = morph_dir if morph_dir else ''
+        self._alternate_morph_dir = alternate_morph_dir if alternate_morph_dir else {}
```
GianlucaFicarelli, line 51: Minor
```suggestion
        self._morph_dir = morph_dir or ''
        self._alternate_morph_dir = alternate_morph_dir or {}
```

bluepysnap/morph.py
```diff
@@ -28,11 +28,16 @@
...
+    'asc': 'neurolucida-asc',
+    'h5': 'h5v1',
+}
+
 
 class MorphHelper:
     """Collection of morphology-related methods."""
 
-    def __init__(self, morph_dir, population):
+    def __init__(self, morph_dir, population, alternate_morph_dir=None):
```
GianlucaFicarelli, line 40: `alternate_morph_dir` should be added to the docstring.
Considering that it's a collection (a dict) and not a single string, a plural name `alternate_morph_dirs` would be clearer.
Or it can be called `alternate_morphologies` that's the same name used in the configuration.

bluepysnap/nodes.py
```diff
@@ -227,15 +231,18 @@ def load_population_data(self, population):
...
 
         Args:
             node_storage (NodeStorage): the node storage containing the node population
             population_name (str): the name of the node population
+            population_config (dict): the config for the population
+
         Returns:
             NodePopulation: A NodePopulation object.
         """
+        self._config = population_config if population_config else {}
```
GianlucaFicarelli, line 245: Minor
```suggestion
        self._config = population_config or {}
```

bluepysnap/bbp.py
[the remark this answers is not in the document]
GianlucaFicarelli: If this code is not actually used (yet), we could move it into a different PR to be completed with the validation code.
Maybe it will be clearer how to organize it, and maybe we'll want to validate both mandatory and optional fields.

bluepysnap/circuit_validation.py
```diff
@@ -381,13 +423,14 @@ def _check_nodes_population(nodes_dict, config):
...
                               format(population_name, nodes_file, missing_datasets))]
             if len(groups) > 1:
-                m_errors = _check_multi_groups(population['node_group_id'],
-                                               population['node_group_index'], population)
-                if len(m_errors) > 0:
-                    return m_errors
+                errors += _check_multi_groups(population['node_group_id'],
+                                              population['node_group_index'], population)
+                if len(errors) > 0:
+                    return errors
```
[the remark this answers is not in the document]
joni-herttuainen: I think the general idea is to return immediately if the error is fatal one.  At least that's the idea I followed. In this case `_check_multi_groups` only has fatal errors so if there's any errors, I'll return it.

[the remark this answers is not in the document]
asanin-epfl: It's a general rule in validation to return if there is a fatal error. Everything is good here. The only problem is that now it validates the spec that is not open-sourced while the project is open-sourced.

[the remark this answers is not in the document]
GianlucaFicarelli: Wouldn't make sense to return all the errors, unless it's not possible to proceed with the validation?
In this way the user would be able to fix all of them at once.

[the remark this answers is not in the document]
asanin-epfl: I didn't check the exact situation here but a fatal error means that validation can't be continued because there is error in Sonata structure that does prevent further reading of data.

bluepysnap/morph.py
joni-herttuainen: I agree. `alternate_morphologies` is best here. I wonder why I decided to go w/ this?

bluepysnap/circuit_validation.py
joni-herttuainen: Sure, no problem.

bluepysnap/circuit_validation.py
joni-herttuainen: Ah, yes. I think I reinforced this when I got the picture that this would be required in the BBP spec. I'll change it. We definitely need a PR to change the validation to only check the BBP spec.

CHANGELOG.rst
```diff
@@ -1,6 +1,19 @@
 Changelog
 =========
 
+Version v0.13.0
+---------------
+
+New Features
+~~~~~~~~~~~~~
+- Population type available in NodePopulation/EdgePopulation
```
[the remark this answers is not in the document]
joni-herttuainen: Yeah, sure.

bluepysnap/circuit_validation.py
joni-herttuainen: Oh, I must've messed it up in the merge. I'll review this.

bluepysnap/circuit_validation.py
joni-herttuainen: Okay, I discussed this with @jdcourcol  and populations indeed is required within BBP. I made a PR to update our documentation.

bluepysnap/circuit_validation.py
joni-herttuainen: I will make another review on that since unpinning pylint would imply changes on files not changed in this PR.

bluepysnap/config.py
joni-herttuainen: They are the list of keys that we expect to have paths (in order to resolve relative paths that don't begin with `./`). I added a comment to explain this.

bluepysnap/circuit_validation.py
```diff
@@ -308,48 +313,81 @@ def _check_rotations():
...
+
+    Args:
+        populations_dict (dict): edge/node populations in config nodes/edges
+        populations_h5 (h5py.Group): nodes/edges group in the h5 file
+        config (str): the name of the h5 file
+
+    Returns:
+        list: List of errors, empty if no errors
+    """
+    not_found = set(populations_dict.keys()) - set(populations_h5.keys())
```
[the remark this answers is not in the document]
joni-herttuainen: fixed

bluepysnap/circuit_validation.py
```diff
@@ -308,48 +313,81 @@ def _check_rotations():
...
+    if not_found:
+        return [fatal('populations not found in {}:\n{}'.format(
+            file_name, ''.join('\t%s\n' % p for p in not_found)))]
```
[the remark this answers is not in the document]
joni-herttuainen: Agreed, changed

bluepysnap/edges.py
```diff
@@ -356,16 +362,18 @@ def _estimate_range_size(func, node_ids, n=3):
...
 
         Args:
             edge_storage (EdgeStorage): the edge storage containing the edge population
             population_name (str): the name of the edge population
+            population_config (dict): the config for the population
 
         Returns:
             EdgePopulation: An EdgePopulation object.
         """
+        self._config = population_config if population_config else {}
```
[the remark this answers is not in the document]
joni-herttuainen: changed

bluepysnap/morph.py
joni-herttuainen: changed

bluepysnap/morph.py
joni-herttuainen: fixed

bluepysnap/nodes.py
joni-herttuainen: fixed

bluepysnap/circuit_validation.py
asanin-epfl: ok

asanin-epfl:
All my issues have been resolved except the code duplication in the fields among Astrocyte, Node, etc. We can keep them as it is now. I'm waiting for Gianluca's confirmation.

GianlucaFicarelli (commented):
> Astrocyte, Node, etc. We can keep them as it is now. I'm waiting for Gianluca's confirmation.

I prefer not to add code that's not used, unless I'm missing something.
Are these public constants that can be used by the users of the library?
If not, what can be the advantage of adding them in this PR?

bluepysnap/config.py
joni-herttuainen: No particular reason other than preference. Set makes sense here, so I might as well use that.

bluepysnap/bbp.py
```diff
@@ -17,7 +17,213 @@
 
 """BBP cell / synapse attribute namespace."""
 
-from bluepysnap.sonata_constants import DYNAMICS_PREFIX, Edge, Node
+from bluepysnap.sonata_constants import DYNAMICS_PREFIX, ConstContainer, Edge, Node
+
+NODE_TYPES = ['biophysical', 'virtual', 'astrocyte', 'single_compartment', 'point_neuron']
+EDGE_TYPES = ['chemical', 'electrical', 'synapse_astrocyte', 'endfoot']
```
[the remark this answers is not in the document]
joni-herttuainen: Yeah. Again, probably due to my preference but set makes sense.

bluepysnap/bbp.py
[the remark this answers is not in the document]
joni-herttuainen: Actually, I think this is used in the ArchNGV PR by @asanin-epfl  and @eleftherioszisis

tests/test_morph.py
```diff
@@ -88,6 +71,37 @@ def test_get_filepath(self):
         with pytest.raises(BluepySnapError):
             self.test_obj.get_filepath([0, 1])
 
+        with pytest.raises(BluepySnapError):
```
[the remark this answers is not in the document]
joni-herttuainen: Agreed

tests/test_morph.py
```diff
@@ -88,6 +71,37 @@ def test_get_filepath(self):
...
+            self.test_obj.get_filepath(0, 'asc')
+
+        with pytest.raises(BluepySnapError):
```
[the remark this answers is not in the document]
joni-herttuainen: Agreed

bluepysnap/edges.py
joni-herttuainen: I tend to agree that if a model type can be specified in the config, it should be validated. Even if, currently, that given type is only used as a return value here.

bluepysnap/bbp.py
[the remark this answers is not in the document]
asanin-epfl: ArchNGV only redirects to those classes. It imports those instances so users of ArchNGV can import them like, it does not use them in its code:
```python
from archngv.core.constants import Astrocyte, Cell, Synapse.
```
https://bbpgitlab.epfl.ch/molsys/ArchNGV/-/merge_requests/21/diffs#106c97ac08ff01bdaf1c3533d4abd91c5b1ac243_21_6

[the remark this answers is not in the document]
joni-herttuainen: I see. If you guys rather not have them defined in this PR, I'm fine not having them. Regarding the model type validation, should I just validate it by comparing to the value in the `*.h5` files then (instead of comparing to what is available)?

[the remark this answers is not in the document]
asanin-epfl: Lets not add them yet then. I didn't understand clearly the proposition with `*.h5` files. Can you please elaborate? May be a small example?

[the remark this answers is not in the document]
joni-herttuainen: Ah, yeah. My idea was to compare it to the `model_type` in the `h5` file, but I just realized that field is not present in the edge files. So I guess, at least for now, we skip the validation, too. Any ideas?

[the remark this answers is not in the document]
asanin-epfl: > Any ideas?

No ideas. I support your proposition that we can skip this validation.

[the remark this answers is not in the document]
joni-herttuainen: As there's no objections, I remove the unused types and don't validate the model type

CHANGELOG.rst
```diff
@@ -1,6 +1,21 @@
...
+Version v0.13.0
+---------------
+
+New Features
+~~~~~~~~~~~~~
+- Node/edge populations are now supported in config
+- Population type available in NodePopulation/EdgePopulation
+- Population config (if given) overwrites the "components" config for that population
+- Alternate morphology directories (.h5, .asc) are now supported
+- Added different edge/node types to bluepysnap/bbp.py
```
asanin-epfl, line 13: Please adjust this line and it is good to go for me.

joni-herttuainen: Right, thanks.

asanin-epfl:
All good for me, @GianlucaFicarelli do you have anything to add?

GianlucaFicarelli:
> All good for me, @GianlucaFicarelli do you have anything to add?

All good for me as well 👍

GianlucaFicarelli (approved):

joni-herttuainen:
thanks guys!

asanin-epfl:
Can you please do a release so ArchNGV can anchor to it?

joni-herttuainen:
Done: https://github.com/BlueBrain/snap/releases/tag/v0.13.0

PR #143, opened 2021-08-05
Fix circuit validation.

asanin-epfl:
Validation of morphologies was skipped when no rotations fields were present.


GianlucaFicarelli (approved):

asanin-epfl:
Thank you

PR #147, opened 2021-10-04
Remove restriction pylint==2.4.4

joni-herttuainen:
Also, fix the linter errors.

### DIFF
--- a/bluepysnap/_plotting.py
+++ b/bluepysnap/_plotting.py
@@ -174,7 +174,7 @@ def spike_raster(filtered_report, y_axis=None, ax=None):  # pragma: no cover
                 ax.set_yticklabels(labels)
                 if len(labels) > 1:
                     ax.set_ylim(-0.5, len(labels) - 0.5)
-            ax.set_ylabel(f"{y_axis}")
+            ax.set_ylabel(y_axis)
 
     ax.scatter(data.index.to_numpy(), data.to_numpy(), s=10, marker='|')
     if len(props["pop_separators"]) > 1:
--- a/bluepysnap/_plotting.py
+++ b/bluepysnap/_plotting.py
@@ -60,7 +60,7 @@ def spikes_firing_rate_histogram(filtered_report, time_binsize=None, ax=None):  
     # pylint: disable=too-many-locals
     plt = _get_pyplot()
     if time_binsize is not None and time_binsize <= 0:
-        raise BluepySnapError("Invalid time_binsize = {}. Should be > 0.".format(time_binsize))
+        raise BluepySnapError(f"Invalid time_binsize = {time_binsize}. Should be > 0.")
 
     spike_report = filtered_report.spike_report
 
@@ -69,7 +69,7 @@ def spikes_firing_rate_histogram(filtered_report, time_binsize=None, ax=None):  
 
     if len(times) == 0:
         raise BluepySnapError("No data to display. You should check your "
-                              "'group' query: {}.".format(spike_report.group))
+                              f"'group' query: {spike_report.group}.")
 
     time_start = np.min(times)
     time_stop = np.max(times)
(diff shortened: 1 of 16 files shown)


bluepysnap/_plotting.py
```diff
@@ -36,7 +36,7 @@ def _get_pyplot():
             "Please pip install as follows:\n"
             "  pip install bluepysnap[plots] --upgrade"
         )
-        raise ImportError(str(e) + "\n\n" + msg)
+        raise ImportError(str(e) + "\n\n" + msg) from e
```
GianlucaFicarelli, line 39:
```suggestion
        raise ImportError(f"{e}\n\n{msg}") from e
```

bluepysnap/_plotting.py
```diff
@@ -174,7 +174,7 @@ def _update_raster_properties():
                 ax.set_yticklabels(labels)
                 if len(labels) > 1:
                     ax.set_ylim(-0.5, len(labels) - 0.5)
-            ax.set_ylabel("{}".format(y_axis))
+            ax.set_ylabel(f"{y_axis}")
```
GianlucaFicarelli, line 177: According to the docstring `y_axis` is already a string (and it cannot be None because of the previous check), so it could be:
```suggestion
            ax.set_ylabel(y_axis)
```

bluepysnap/node_sets.py
```diff
@@ -75,9 +75,9 @@ def _resolve_set(content, resolved, node_set_name):
     # keep the content intact
     set_value = deepcopy(content.get(node_set_name))
     if set_value is None:
-        raise BluepySnapError("Missing node_set: '{}'".format(node_set_name))
+        raise BluepySnapError(f"Missing node_set: '{node_set_name}'")
     if not isinstance(set_value, (Mapping, list)) or not set_value:
-        raise BluepySnapError("Ambiguous node_set: '{}'".format({node_set_name: set_value}))
+        raise BluepySnapError(f"Ambiguous node_set: '{dict(node_set_name=set_value)}'")
```
GianlucaFicarelli, line 80: A possible alternative (removing `''` that don't seem useful to wrap a dict, and using `{}` instead of `dict`):
```suggestion
        raise BluepySnapError(f"Ambiguous node_set: { {node_set_name: set_value} }")
```

bluepysnap/spike_report.py
```diff
@@ -225,7 +225,7 @@ def log(self):
         path = Path(self.config["output_dir"]) / self.config["log_file"]
         if not path.exists():
             raise BluepySnapError("Cannot find the log file for the spike report.")
-        yield open(str(path), "r")
+        yield open(str(path), "r", encoding='utf-8')
```
GianlucaFicarelli, line 228: Unrelated to this PR, but it could use the path open method:
```suggestion
        yield path.open("r", encoding='utf-8')
```

bluepysnap/_plotting.py
mgeplf, line 39: this `str(e)` still needed?

GianlucaFicarelli (commented):
LGTM, I just added a few minor suggestions.

bluepysnap/_plotting.py
joni-herttuainen: Good point, I think reporting the message with the available stack trace (w/ `from e`) is redundant. I'll remove it.

bluepysnap/_plotting.py
joni-herttuainen: Nice catch!

bluepysnap/spike_report.py
joni-herttuainen: Cool. Might as well fix it.

GianlucaFicarelli (approved):

PR #149, opened 2021-11-04
Bump click requirement

GianlucaFicarelli:


lguyot:
I would be happy to approve very soon. Any word about the context? Is it now safe to remove the version cap?

GianlucaFicarelli:
> I would be happy to approve very soon. Any word about the context? Is it now safe to remove the version cap?

Sure, there is no hurry to approve. I propose to remove the upper limit for `click` because:
- `bluepysnap` is mainly a library. Click is used only for the CLI `bluepysnap validate`, and only for some basic cases that are unlikely to change (even though I cannot know for sure):
```python
@cli.command('validate', short_help='Validate Sonata circuit')
@click.argument('config_file', type=click.Path(exists=True, file_okay=True, dir_okay=False))
```
and
```
click.style('No Error: Success.', fg=colors[Error.INFO])
```
- click 8 has been released, but because of this limitation, it cannot be used by any package depending on snap.
- the issue in spack would have been avoided, but the root cause of that issue is only the discrepancy between the version of click specified in spack and setup.py.

lguyot (approved):
LGTM. Thanks for your explanations!

joni-herttuainen (approved):

PR #151, opened 2022-02-03
Add python 3.9 tests

GianlucaFicarelli:
- Add python 3.9 tests.
- Ensure that ids in frame reports are always np.int64 even when using libsonata 0.1.10 (see NSETM-1766).
- Fix deprecation warnings.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,10 +4,13 @@ Changelog
 Version v0.13.1
 ---------------
 
+Improvements
+~~~~~~~~~~~~
+- Add python 3.9 tests
+
 Removed
 ~~~~~~~
 - Drop python 3.6 support
-- Add python 3.9 tests
 
 
 Version v0.13.0
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,8 @@ Version v0.13.1
 
 Removed
 ~~~~~~~
-- Drop python3.6 support
+- Drop python 3.6 support
+- Add python 3.9 tests
 
 
 Version v0.13.0
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,11 +6,15 @@ Version v0.13.1
 
 Improvements
 ~~~~~~~~~~~~
-- Add python 3.9 tests
+- Add python 3.9 tests.
+
+Bug Fixes
+~~~~~~~~~
+- Ensure that ids in frame reports are always np.int64 even when using libsonata 0.1.10.
 
 Removed
 ~~~~~~~
-- Drop python 3.6 support
+- Drop python 3.6 support.
 
 
 Version v0.13.0
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ Improvements
 Bug Fixes
 ~~~~~~~~~
 - Ensure that ids in frame reports are always np.int64 even when using libsonata 0.1.10.
+- Fix deprecation warnings.
 
 Removed
 ~~~~~~~
--- a/bluepysnap/_plotting.py
+++ b/bluepysnap/_plotting.py
@@ -22,7 +22,7 @@ from more_itertools import roundrobin
 
 from bluepysnap.exceptions import BluepySnapError
 from bluepysnap.sonata_constants import Node
-
+from bluepysnap.utils import IDS_DTYPE
 
 L = logging.getLogger(__name__)
 
@@ -133,8 +133,9 @@ def spike_raster(filtered_report, y_axis=None, ax=None):  # pragma: no cover
 
     report = filtered_report.report
 
-    dtype = spike_report[population_names[0]].nodes.property_dtypes[y_axis] if y_axis else None
-    if dtype and pd.api.types.is_categorical_dtype(dtype):
+    # use np.int64 if displaying node_ids
+    dtype = spike_report[population_names[0]].nodes.property_dtypes[y_axis] if y_axis else IDS_DTYPE
+    if pd.api.types.is_categorical_dtype(dtype):
         # this is to prevent the problems when concatenating categoricals with unknown categories
         dtype = str
     data = pd.Series(index=report.index, dtype=dtype)
--- a/bluepysnap/edges.py
+++ b/bluepysnap/edges.py
@@ -102,15 +102,19 @@ class Edges(NetworkObject, metaclass=AbstractDocSubstitutionMeta,
         if edge_ids is None:
             raise BluepySnapError("You need to set edge_ids in get.")
         if properties is None:
-            Deprecate.warn("Returning ids with get/properties will be removed in 1.0.0."
-                           "Please use Edges.ids() instead.")
+            Deprecate.warn(
+                "Returning ids with get/properties is deprecated and will be removed in 1.0.0. "
+                "Please use Edges.ids instead."
+            )
             return edge_ids
         return super().get(edge_ids, properties)
 
     def properties(self, edge_ids, properties):
         """Doc is overridden below."""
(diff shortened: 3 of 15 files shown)


joni-herttuainen (approved):

CHANGELOG.rst
```diff
@@ -6,7 +6,8 @@ Version v0.13.1
 
 Removed
 ~~~~~~~
-- Drop python3.6 support
+- Drop python 3.6 support
+- Add python 3.9 tests
```
joni-herttuainen, line 10: minor: should this rather be in `Improvements`?

GianlucaFicarelli: Right!

joni-herttuainen: LGTM

eleftherioszisis (approved):

bluepysnap/_plotting.py
```diff
@@ -133,8 +133,9 @@ def _update_raster_properties():
 
     report = filtered_report.report
 
-    dtype = spike_report[population_names[0]].nodes.property_dtypes[y_axis] if y_axis else None
-    if dtype and pd.api.types.is_categorical_dtype(dtype):
+    # use np.int64 if displaying node_ids
+    dtype = spike_report[population_names[0]].nodes.property_dtypes[y_axis] if y_axis else IDS_DTYPE
```
GianlucaFicarelli, line 137: Using `None` as dtype here was causing bluepy functional tests to fail because of this warning from pandas:
> FutureWarning: The default dtype for empty Series will be 'object' instead of 'float64' in a future version. Specify a dtype explicitly to silence this warning.

tests/test_nodes.py
```diff
@@ -656,7 +656,9 @@ def test_get(self):
         )
 
         assert _call("Node0_L6_Y", properties=[Cell.X, Cell.MTYPE, Cell.LAYER]).empty
-        assert _call(1, properties={Cell.MTYPE}).tolist() == ["L6_Y"]
+        assert _call(1, properties=[Cell.MTYPE]).tolist() == ["L6_Y"]
+        assert _call([1], properties=Cell.MTYPE).tolist() == ["L6_Y"]
+        assert _call([1, 2], properties=Cell.MTYPE).tolist() == ["L6_Y", "L6_Y"]
```
GianlucaFicarelli, line 661: Passing a dict as `properties` causes a warning in pandas:
> FutureWarning: Passing a set as an indexer is deprecated and will raise in a future version. Use a list instead.

I removed the previous test because the `properties` parameter in the `get` method is expected to be a list or a str (as explained in the dostring), not a dict. If a dict is used, it still works with the current version of pandas, but a warning is printed (by pandas itself).

I added a test passing (list of ids, properties=str).

bluepysnap/nodes.py
```diff
@@ -545,14 +545,18 @@ def get(self, group=None, properties=None):
...
             value/pandas.Series/pandas.DataFrame:
-                If single node ID is passed as ``group`` and single property as properties returns
-                a single value. If single node ID is passed as ``group`` and list as property
-                returns a pandas Series. Otherwise return a pandas DataFrame indexed by node IDs.
+                If a single node ID is passed as ``group`` and a single property as ``properties``,
+                returns a single value.
+                If a single node ID is passed as ``group`` and a list as ``properties``,
+                returns a pandas Series.
+                If something different from a single node ID is passed as ``group`` and a single
+                property as ``properties``, returns a pandas Series.
```
joni-herttuainen, line 558: I read this sentence as:
`len(group) !=1 AND len(properties) != 1`
while I believe the meaning behind the sentence is:
`len(group) !=1 AND len(properties) == 1`

Maybe 
"...single node ID is passed as ``group`` and a single property is passed as ``properties``..."
would do the trick?

I don't know. I'm not a native speaker either so... 🤷‍♂️ 

Also, I'd use: `If anything else than a single node ID...` but I'm still not a native speaker 🤣

bluepysnap/_plotting.py
joni-herttuainen: makes sense

tests/test_nodes.py
joni-herttuainen: Ok, cool 👍

bluepysnap/edges.py
```diff
@@ -614,8 +620,10 @@ def get(self, edge_ids, properties):
 
     def properties(self, edge_ids, properties):
         """Doc is overridden below."""
-        Deprecate.warn("EdgePopulation.properties function will be deprecated in 1.0.0. Please use "
-                       "EdgePopulation.get instead.")
+        Deprecate.warn(
+            "EdgePopulation.properties function is deprecated and will be removed in 1.0.0. "
+            "Please use EdgePopulation.get instead."
+        )
```
joni-herttuainen, line 626: Regarding these, these are only style changes and probably should not be a part of this PR, right?

That being said, the more I've worked on snap lately, the more I have wanted to just (in the lack of a better word) blackify it and let `black` do all the heavy lifting regarding the styling. What do you think?

bluepysnap/nodes.py
GianlucaFicarelli: Yeah, I'll rephrase it. Note that if you pass `1` instead of `[1]` as `group` the results are different, so I can add an example for each case to make it clearer.

bluepysnap/edges.py
GianlucaFicarelli: This is not only a format change, but I wanted to fix the message: the message was saying that "the function will be deprecated in 1.0.0", but the function is already deprecated in the current version. In 1.0.0 it will be removed (if not, the message should be fixed in a different way).
Even though the changes are not related to python 3.9, I did it in this PR while updating to python 3.9 because I was fixing also the warnings caused by the new version of panda.

About black, I'm in with it, also considering that it's now considered not beta anymore. We can discuss it directly 👍

joni-herttuainen: oh, you're absolutely correct. I missed that as it was not inline-highlighted. my bad. 🖐️
Okay, let's keep it here, then. 

Yeah, if you ask me, black is the new... eh, black? 😅

eleftherioszisis (approved):

bluepysnap/frame_report.py
```diff
@@ -106,9 +106,11 @@ def get(self, group=None, t_start=None, t_stop=None):
         if len(view.ids) == 0:
             return pd.DataFrame()
 
-        res = pd.DataFrame(data=view.data,
-                           columns=pd.MultiIndex.from_arrays(np.asarray(view.ids).T),
-                           index=view.times).sort_index(axis=1)
+        res = pd.DataFrame(
+            data=view.data,
+            columns=pd.MultiIndex.from_arrays(np.asarray(view.ids, dtype=np.int64).T),
```
mgeplf, line 111: should n`p.int64` be `IDS_DTYPE`?

GianlucaFicarelli: From what I see `IDS_DTYPE` is used for nodes and edges ids.
In this case the numpy array used for the columns is an array with shape (N, 2), where the first column represents the node ids and the second the section ids.
Considering that they are still ids, we could use `IDS_DTYPE` also in this case, or call directly `utils.ensure_ids`, what do you think?

mgeplf: I think only the first column needs to be set to IDS_DTYPE, so which-ever way makes that cleanest.

I think adding a comment explaining why we're moving from unsigned -> signed is worthwhile though - just a link to the numpy ticket would do, imo.

mgeplf (approved):

bluepysnap/frame_report.py
GianlucaFicarelli: Discussed and agreed to make int64 both cell ids and section ids, to avoid any issue with automatic conversions and for consistency with snap when using libsonata<0.1.10.

PR #154, opened 2022-03-24
Using libsonata to cover more of the use cases

joni-herttuainen:
* Use cases are working more or less the same as they used to
* `{Edge,Node}Storage` classes were removed
* If there are issues in the config file, the opening of a config file is halted
* manifest parsing needs to be kept due to the validation code (might be solved in #155)


bluepysnap/config.py
```diff
@@ -132,6 +131,40 @@ def resolve(self):
...
+    def from_simulation_config(cls, config_path):
+        """Instantiate the config class from simulation configuration."""
+        return cls(config_path, libsonata.SimulationConfig)
+
+    def to_dict(self):
+        """Return the configuration as a dict with absolute paths."""
+        if hasattr(self._libsonata, "expanded_json"):
+            config = json.loads(self._libsonata.expanded_json)
+        else:
+            config = json.loads(self._libsonata.json)
```
eleftherioszisis, line 168: Out of curiosity, what is the diff between the two?

bluepysnap/network.py
```diff
@@ -38,10 +38,6 @@ def _get_populations(self, cls, config):
         for file_config in config:
             storage = cls(file_config, self._circuit)
             for population in storage.population_names:  # pylint: disable=not-an-iterable
-                if population in res:
-                    raise BluepySnapError(
-                        f"Duplicated {self.__class__.__name__} population: '{population}'"
-                    )
```
eleftherioszisis, line 44: Should this be allowed?

eleftherioszisis:
Do you have any insight if there could be edge cases where the libsonata replacement could break backward compatibility with these changes?

bluepysnap/config.py
joni-herttuainen: `expanded_json` is currently only available in `libsonata.CircuitConfig` and it expands the manifest values in the config, whereas `json` is just the raw json data as it is in the (in this case simulation) configuration file.

bluepysnap/network.py
joni-herttuainen: No, not at all, but if that is the case, the libsonata config parser raises an error so this code will never be executed.

bluepysnap/config.py
```diff
@@ -132,6 +131,45 @@ def resolve(self):
...
+        Args:
+            config (str): Path to the configuration file
+            config_class (class): libsonata class corresponding to the configuration file, either
+                libsonata.CircuitConfig or libsonata.SimulationConfig
+        """
+        self._config_dir = str(Path(config).parent.absolute())
+        self._libsonata = config_class.from_file(config)
+
+    @property
+    def instance(self):
```
eleftherioszisis, line 154: I would suggest `to_libsonata` here as well.

bluepysnap/edges.py
```diff
@@ -903,6 +846,10 @@ def iter_connections(
             return map(omit_edge_count, it)
 
     @property
-    def h5_filepath(self):
+    def h5_filepath(self):  # pylint: disable=inconsistent-return-statements
         """Get the H5 edges file associated with population."""
-        return self._edge_storage.h5_filepath
+        for edge_conf in self.circuit.config["networks"]["edges"]:
+            h5_filepath = edge_conf["edges_file"]
+            # set(...) as pylint seems to think {}.population_names doesn't return a set
```
eleftherioszisis, line 853: If pylint is wrong here, disabling the error would be better, no?

bluepysnap/edges.py
```diff
@@ -47,8 +46,14 @@ def __init__(self, circuit):  # pylint: disable=useless-super-delegation
...
-    def _collect_populations(self):
-        return self._get_populations(EdgeStorage, self._config["networks"]["edges"])
+    @property
+    def _population_class(self):
+        return EdgePopulation
+
+    @cached_property
+    def population_names(self):
+        """Defines all sorted edge population names from the Circuit."""
+        return sorted(self._circuit.libsonata.edge_populations)
```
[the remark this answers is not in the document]
eleftherioszisis: Just checked libsonata's CircuitConfig and realized why this is needed. It is a set.

bluepysnap/edges.py
joni-herttuainen: Yeah, fixed

bluepysnap/bbp.py
```diff
@@ -19,7 +19,7 @@
 
 from bluepysnap.sonata_constants import DYNAMICS_PREFIX, Edge, Node
 
-NODE_TYPES = {"biophysical", "virtual", "astrocyte", "single_compartment", "point_neuron"}
+NODE_TYPES = {"biophysical", "virtual", "astrocyte", "single_compartment"}
```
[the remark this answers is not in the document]
eleftherioszisis: okk, I should remove it as a possibility for writing this pop to sonata config in circuit-build.

bluepysnap/config.py
mgeplf: I have fixed this; both `CircuitConfig` and `SimulationConfig` should be the same now:
https://github.com/BlueBrain/libsonata/pull/214

bluepysnap/config.py
```diff
@@ -37,32 +39,34 @@
     "edges_type_file",
     "nodes_type_file",
     "node_sets_file",
+    "output_dir",
+    "network",
+    "mechanisms_dir",
```
eleftherioszisis, line 44: This is probably not relevant to this PR:
If we also consider the ngv keys (I noticed vasculature is here too), there are a couple missing, such as:
* microdomains_file
* endfeet_meshes_file (this should replace the 'end_feet_areas')

bluepysnap/config.py
joni-herttuainen: I'll add this `SimulationConfig.expanded_json`, too.

Due to my short-sightedness on the matter, though, I just realized I can not remove the manifest resolution in the `Parse` class as it is needed by the validation code. :man_facepalming: 

However, when we go forward with the schema based validation for the config file in #155 , I'd _assume_ this would then no longer be needed.

bluepysnap/config.py
```diff
@@ -37,32 +39,33 @@
...
 
     See Also:
         https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#network_config
     """
 
-    def __init__(self, config):
-        """Initializes a Config object from a path to the actual config.
+    # TODO: Can be simplified by a great deal after libsonata SimulationConfig parses the manifest
+
+    def __init__(self, config, configdir):
```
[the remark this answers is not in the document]
joni-herttuainen: Now I remember this, too: I can not have the `Parse` class rely on `libsonata` as the validation code relies on `Parse` and `libsonata` does not support iterating over different errors in the config.

So my suggestion is to go with this as of now, and when the config is validated by schemas (#155), we can then re-evaluate how'd we want to proceed with the further validation and how much of the code we can clean up.

bluepysnap/edges.py
```diff
@@ -903,6 +846,10 @@ def iter_connections(
             return map(omit_edge_count, it)
 
     @property
-    def h5_filepath(self):
+    def h5_filepath(self):  # pylint: disable=inconsistent-return-statements
```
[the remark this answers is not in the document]
joni-herttuainen: all the node and edge populations should be part of a h5 file, so error it is.

bluepysnap/edges.py
```diff
@@ -902,7 +845,12 @@ def iter_connections(
             omit_edge_count = lambda x: x[:2]
             return map(omit_edge_count, it)
 
-    @property
+    @cached_property
     def h5_filepath(self):
         """Get the H5 edges file associated with population."""
-        return self._edge_storage.h5_filepath
+        for edge_conf in self._circuit.config["networks"]["edges"]:
```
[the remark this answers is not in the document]
joni-herttuainen: Makes sense -> changed

bluepysnap/edges.py
```diff
@@ -311,63 +316,6 @@ def iter_connections(
                 yield from self._omit_edge_count(it, source_pop, target_pop)
 
 
-class EdgeStorage:
```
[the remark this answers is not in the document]
joni-herttuainen: It seems it's the much the same in `bluepy`: to open populations. In the case of node populations, it's written in a way that the circuit wouldn't need to be opened. We probably need to overwrite some of the methods but I don't see major issues.

tests/test_edges.py
```diff
@@ -805,13 +754,23 @@ def test_basic(self):
                 test_module.DYNAMICS_PREFIX + "param1",
             ]
         )
-        assert self.test_obj.type == DEFAULT_EDGE_TYPE
+        # TODO: change to DEFAULT_EDGE_TYPE when fixed in libsonata (= "chemical")
```
mgeplf, line 757: is this fixed?

tests/test_config.py
```diff
@@ -92,83 +99,60 @@ def test_parse():
         assert actual["path"] == str(Path(config_path.parent / "./path").resolve())
 
 
+# TODO: remove this when SimulationConfig has manifest parsing
```
mgeplf, line 102: is this fixed in libsonata?

tests/test_edges.py
joni-herttuainen: It seems this is `chemical_synapse` in `libsonata`: https://github.com/BlueBrain/libsonata/blob/master/src/config.cpp#L744

tests/test_config.py
joni-herttuainen: Yeah, this actually currently depends on when/if we can remove the Parser in config (as it is still used by circuit_validation)
-> removed comment

tests/test_edges.py
mgeplf: huh; the doc says it's also `chemical`: https://sonata-extension.readthedocs.io/en/latest/sonata_config.html#populations

but I swear it's inconsistent somewhere else - let me look

mgeplf (approved):

mgeplf:
:tada:

PR #155, opened 2022-04-05
Validator

mgeplf:


bluepysnap/circuit_validation.py
```diff
@@ -740,11 +740,117 @@ def validate(config_file, bbp_check=False):
...
+        elif isinstance(value, h5py.Dataset):
+            properties[key] = {'datatype': get_dataset_dtype(value)}
+        else:
+            properties[key] = {}
+
+        attrs = {k: v for k, v in value.attrs.items()}
+        if attrs:
+            properties[key]['attributes'] = attrs
+
+    # Maybe check for the existence of @library and do the translation to actual values here?
```
mgeplf, line 791: yeah, that would be a good idea; perhaps it's worth also noting that they come from `@library`, in case we want to verify that too

bluepysnap/circuit_validation.py
```diff
@@ -740,11 +740,117 @@ def validate(config_file, bbp_check=False):
...
+        import jsonschema
+        validator = jsonschema.validators.Draft202012Validator(schema)
+        # currently just printing the errors but the idea would be to have something like
+        # [heavily pseudo code]
+        # errors[field] = e.type
+        # if e.type != 'warning':
+        #   failed_fields.append(e.absolute_path)
+        # ... later ...
+        # validate_data(data_file, file_structure - failed_fields)
+        for e in validator.iter_errors(dict_):
```
mgeplf, line 821: yeah, agree about adding the errors to a list that returned to the caller is the best way

joni-herttuainen:
Apologies for, yet again, huge diff.

### Schemas
First of all, I moved schema code to `bluepysnap/schemas/schemas.py`. I also applied "schema inheritance" as much as it can be called that. I guess schema re-use would be a better term.

In `bluepysnap/schemas/definitions`, there are the base node and edge file schemas as well as the datatype definitions. 

The population type specific definitions are in `bluepysnap/schemas/{node,edge}/*` (all definitions are not done yet).  They are created to "inherit" the base class and the type definitions. This is done in `schemas.py:_parse_schema`. The circuit config schema definition is currently in `bluepysnap/schemas/circuit.yaml`.

Please note that none of the schemas contain descriptions on any of the properties yet (for creating docs).

I haven't worked on the test cases yet, but I've been running tests on some of the circuit configs in GPFS.

A runnable version with all the other circuit validation stuff still intact is in commit e8b6814. This can be run with
```bash
bluepysnap validate /gpfs/bbp.cscs.ch/project/proj12/NSE/bluepysnap-functional-tests/sonata/circuit_sonata.json --bbp-check
```
and it will print errors and exit after schema validation (does not continue the normal validation).

### Validation rework
I reworked the circuit validation logic in the latest commit (2aeb9dd). It is a big diff mostly due to me being able to remove a lot of code for two reasons. First of all, missing mandatory fields is covered with schemas and secondly, I removed all the non-BBP validations.

Another reason is that I changed the logic how validation operates. 

#### New logic
While the all logic was trying to cover everything in the circuit config and in the node files and halt on `fatal` errors, the new logic is much more forgiving. First of all, the validation is driven by the circuit configuration file. What that means, is basically that I only check stuff that is listed in the configuration. If the nodes file had 5 population but only one was listed in config, only that one is validated. If no populations are listed, no populations are validated. If population type in the config is different from that in the h5 file, (also checked later on) the schema based on the type in the circuit config is used.

The schema is still checked for the whole nodes/edges file at the time, though. 

Secondly, I have modified the code to validate as far as it can get. Even if there are some issues and errors during , it continues with other checks. 

### Points of discussion
* only bbp spec?
  * I recall we discussed abandoning validation of the official spec, as we don't even support it in the code
* circuit config driven validation
  * should we validate the whole h5 file if multiple populations present
* to print or to return errors
  * currently returned but I fail to see the benefit and would rather just print
* multipopulation (per h5 file) support?

@jdcourcol, @mgeplf, @GianlucaFicarelli

bluepysnap/schemas/definitions/node.yaml
```diff
@@ -0,0 +1,76 @@
...
+    maxProperties: 1
+    minProperties: 1
+    type: object
+    patternProperties:
+      "":
+        type: object
+        required:
+          - "0"
+          - node_type_id
+        additionalProperties: false
```
mgeplf, line 14: I wonder if this is the case; we may have to figure out a way to be permissive in some cases.

bluepysnap/schemas/node/biophysical.yaml
```diff
@@ -0,0 +1,25 @@
+title: Nodes - biophysical
+description: schema for biophysical node types
+properties:
+  nodes:
+    allOf: [{ $ref: "#/$node_file_defs/nodes" }]
+    patternProperties:
+      "":
```
mgeplf, line 7: why is the empty string used here?

mgeplf (commented):
none of the tests had to change?

bluepysnap/circuit_validation.py
```diff
@@ -439,170 +282,47 @@ def _check_nodes_group(group_df, group, config, population):
...
-            groups = _get_population_groups(population)
-
-            if len(groups) > 1:
-                required_datasets.update(
-                    (
-                        "node_group_id",
-                        "node_group_index",
-                    )
-                )
+        nodes_h5 = h5f.get("nodes")
```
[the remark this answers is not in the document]
joni-herttuainen: good point, I used the `h5.get('stuff')` in this, more of a still POC, PR to work around the need for the
```python
if 'stuff' not in h5:
    return []
```
I will clean everything for the next iteration if the change of validation methods is the direction we want to proceed.

bluepysnap/circuit_validation.py
```diff
@@ -11,10 +11,12 @@
 import numpy as np
 import pandas as pd
 
+import bluepysnap.schemas as schemas
```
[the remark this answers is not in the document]
joni-herttuainen: LOL, good point.

bluepysnap/schemas/definitions/node.yaml
joni-herttuainen: Yeah, according to my recent research jsonschema stuff, we should abstain from using `additionalProperties: false` unless it's a clear case. I need to review the cases I used it.

bluepysnap/schemas/node/biophysical.yaml
joni-herttuainen: It's used as a wild card for the population name. Currently, any population name would suffice. We could create some rules for population names, such as 
* accept only `a-Z`, `0-9` and `_`

but those would need to be agreed upon.

joni-herttuainen (commented):
About the need to change tests:
If you mean tests when validating: no not really, all the checks are more or less the same, it's mainly just what to do when they fail and not even testing if stuff is missing

If you mean the tests in `bluepysnap/tests/test_circuit_validation.py`: these obviously need to be changed. I did not change them yet as I wanted to get the general idea open for discussion first. If you feel that these changes are in the direction we wish to go, I will start cleaning stuff up as well as fixing and adding tests.

mgeplf:
> only bbp spec?
> I recall we discussed abandoning validation of the official spec, as we don't even support it in the code
> circuit config driven validation

Let's only cover the BBP spec - at some point in the future, if it's neaded, we can get fancy, but let's keep it simple for now.

> should we validate the whole h5 file if multiple populations present to print or to return errors

I think just validating what's in the config is best.

> currently returned but I fail to see the benefit and would rather just print

It's fine how it is; possible less interactive, but w/ the 'skip-slow', it's bearable.

bluepysnap/circuit_validation.py
```diff
@@ -283,23 +285,17 @@ def _check_nodes_group(group_df, group, population, population_name):
...
-        return errors
-
-    if group_df["model_type"][0] != population["type"]:
-        message = (
-            f"Population '{population_name}' type mismatch: "
-            f"'{group_df['model_type'][0]}' (nodes_file), "
-            f"'{population['type']}' (config)'"
-        )
-        errors.append(Error(Error.WARNING, message))
+    if "model_type" in group_df:
```
mgeplf, line 288: these `if`s can be combined, right?

bluepysnap/circuit_validation.py
```diff
@@ -316,44 +312,44 @@ def validate_node_population(nodes_file, population_dict, name):
         list: List of errors, empty if no errors
     """
     with h5py.File(nodes_file, "r") as h5f:
-        nodes_h5 = h5f.get("nodes")
-
-        if not nodes_h5 or len(nodes_h5) == 0:
+        if "nodes" not in h5f or len(h5f["nodes"]) == 0:
```
mgeplf, line 315: I'd say it's an error if the nodes file isn't actually a node file

bluepysnap/circuit_validation.py
```diff
@@ -316,44 +312,44 @@ def validate_node_population(nodes_file, population_dict, name):
...
+        # special case in which there are populations but not the one expected
+        if name not in h5f["nodes"]:
             return [fatal(f"population '{name}' not found in {nodes_file}")]
 
-        population = nodes_h5[name]
-        group = population.get("0")
+        population = h5f[f"nodes/{name}"]
 
-        if group and population.get("node_type_id"):
+        if "0" in population and "node_type_id" in population:
```
mgeplf, line 324: not having "0" is also an error, or is that handled elsewhere?

bluepysnap/circuit_validation.py
```diff
@@ -316,44 +312,44 @@ def validate_node_population(nodes_file, population_dict, name):
...
-        node_ids = node_population["node_id"][:]
-    else:
-        grp = node_population.get("0")
-        if grp:
-            for attr in grp:
-                if isinstance(grp[attr], h5py.Dataset):
-                    node_ids = np.arange(len(grp[attr]))
-                    break
-    return node_ids
+    node_population = nodes_h5.get(f"nodes/{population_name}")
```
mgeplf, line 343: I still find not using  `nodes_h5["nodes"][population_name]` jarring

bluepysnap/schemas/node/biophysical.yaml
mgeplf: icic, can you add a comment (including in the other cases where an empty pattern is used)?

tests/test_circuit_validation.py
```diff
@@ -1,37 +1,34 @@
 try:
-    from unittest.mock import patch
+    from unittest.mock import Mock, patch
```
mgeplf, line 2: can we just use the unittest ones now? or is our lower bound python still too old?

tests/test_circuit_validation.py
```diff
@@ -531,253 +385,59 @@ def test_no_template_library_files():
...
-        errors = validate(str(config_copy_path))
-        assert errors == {
-            Error(
-                Error.FATAL,
-                "Group default/0 in file {} should have ids up to {}".format(edges_file, 12),
-            )
-        }
-
-
-def test_edge_population_wrong_group_index_multi_group():
```
mgeplf, line 751: are the tests that are deleted covered by other things or not relevant any more?

bluepysnap/circuit_validation.py
joni-herttuainen: I will change this

bluepysnap/circuit_validation.py
joni-herttuainen: Yep, this is handled in schema validation.

bluepysnap/circuit_validation.py
joni-herttuainen: True, I'll change it.

tests/test_circuit_validation.py
joni-herttuainen: I'll look into it

bluepysnap/circuit_validation.py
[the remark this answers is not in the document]
mgeplf: is this done?

mgeplf:
LGTM

bluepysnap/circuit_validation.py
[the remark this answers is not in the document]
joni-herttuainen: yeah I left the `get` style in some places where it made more sense. E.g.,
```python
    source = h5.get(f"edges/{edge_population}/source_node_id")                                                                        
    source_population = source.attrs.get("node_population") if source else None
```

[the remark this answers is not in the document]
mgeplf: ok

joni-herttuainen (approved):
Need to approve myself as @mgeplf created the original PR.

PR #158, opened 2022-05-30
make h5 path accessible from (Node|Edge)Population

mgeplf:


bluepysnap/edges.py
```diff
@@ -901,3 +901,8 @@ def iter_connections(
         else:
             omit_edge_count = lambda x: x[:2]
             return map(omit_edge_count, it)
+
+    @property
+    def h5_filepath(self):
```
eleftherioszisis, line 906: I wonder, is the h5 prefix needed here?

mgeplf: I'm matching the names of what is called; uniformity is good, I think.  If we ever want to do the `CSV` stuff, (shiver), it's safer to have them named.

joni-herttuainen (approved):

PR #138, opened 2021-06-09
NSETM-1298 entity api POC

GianlucaFicarelli:
This PR is a draft to keep track of the changes done in NSETM-1298.
It should not be merged until it's clarified if it can be useful, or eventually deleted together with the branch.

PR #164, opened 2022-08-17
Added Nexus notebooks

joni-herttuainen:
Separate PR for nexus notebooks. Related to #138


joni-herttuainen:
Should not be merged: needs to be moved to bluepyentity project

PR #165, opened 2022-08-30
Changed Simulation and Reports to take more advantage of libsonata

joni-herttuainen:


bluepysnap/frame_report.py
```diff
@@ -39,9 +39,9 @@ def _collect_population_reports(frame_report, cls):
 
 
 def _get_reader(reader_report, cls):
-    path = reader_report.simulation.config["output"]["output_dir"]
-    ext = FORMAT_TO_EXT[reader_report.config.get("format", "HDF5")]
-    file_name = reader_report.config.get("file_name", reader_report.name) + ext
+    path = reader_report.simulation.output.output_dir
+    file_name = reader_report.config.get("file_name", reader_report.name) + REPORT_EXTENSION
+    # file_name = reader_report.to_libsonata.file_name + REPORT_EXTENSION
```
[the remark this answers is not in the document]
mgeplf: seems like a libsonata bug? can you open an issue in that repo?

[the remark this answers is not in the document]
joni-herttuainen: Sure thing: https://github.com/BlueBrain/libsonata/issues/221

bluepysnap/simulation.py
```diff
@@ -67,29 +75,34 @@ def circuit(self):
...
     def run(self):
         """Access to the complete run dictionary for this simulation."""
-        return self.config["run"]
+        return self.to_libsonata.run
 
     @property
     def time_start(self):
         """Returns the starting time of the simulation."""
-        return self.run.get("tstart", 0)
+        return getattr(self.run, "tstart", 0)
```
GianlucaFicarelli, line 95: I don't see `tstart` in https://github.com/BlueBrain/libsonata/blob/e0cdaf707feae298d432a11742d2620721a1ae10/python/bindings.cpp#L530
or https://github.com/BlueBrain/libsonata/blob/e0cdaf707feae298d432a11742d2620721a1ae10/include/bbp/sonata/config.h#L192
and it's not even defined at https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#run
🤔

bluepysnap/simulation.py
```diff
@@ -100,19 +113,19 @@ def time_units(self):
     @property
     def conditions(self):
         """Access to the conditions dictionary for this simulation."""
-        return self.config.get("conditions", {})
+        return getattr(self.to_libsonata, "conditions", None)
```
GianlucaFicarelli, line 116: With this change, `conditions` would return `None` instead of `{}` when the attribute is missing.
Is it intentional? In theory it could break code using it.

tests/test_simulation.py
```diff
@@ -80,16 +81,18 @@ def test_no_network_config():
         with edit_config(config_path) as config:
             config.pop("network")
 
+        os.remove(config_path.parent / "circuit_config.json")
         simulation = test_module.Simulation(config_path)
 
-        with pytest.raises(BluepySnapError, match="No 'network' set"):
+        with pytest.raises(BluepySnapError, match="'network' file not found"):
```
GianlucaFicarelli, line 87: The test is now different and should be renamed from `test_no_network_config` to something like `test_network_file_not_found`.
However, maybe the old test `test_no_network_config` could be kept to verify also that case.

tests/test_simulation.py
```diff
@@ -80,16 +81,18 @@ def test_no_network_config():
...
             simulation.circuit
 
 
 def test_no_node_set():
     with copy_test_data(config="simulation_config.json") as (_, config_path):
         with edit_config(config_path) as config:
             config.pop("node_sets_file")
+            os.remove(config_path.parent / "circuit_config.json")
```
GianlucaFicarelli, line 95: Why the file `circuit_config.json` needs to be removed to make the test succeed?
Isn't it related to the configuration only?

GianlucaFicarelli:
The new changes can be mentioned more explicitly in the changelog.

bluepysnap/simulation.py
```diff
@@ -67,29 +75,34 @@ def circuit(self):
         """Access to the circuit used for the simulation."""
         from bluepysnap.circuit import Circuit
 
-        if "network" not in self.config:
-            raise BluepySnapError("No 'network' set in the simulation/global config file.")
-        return Circuit(self.config["network"])
+        if not Path(self.to_libsonata.network).is_file():
```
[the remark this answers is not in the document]
joni-herttuainen: [`libsonata`](https://github.com/BlueBrain/libsonata/blob/master/src/config.cpp#L933) defaults `network` to `{base_path}/circuit_config.json` if not found.


I don't know if it'd be better to remove the default from `libsonata` and just do
```python
return getattr(self.to_libsonata, 'network', None)
```
or revert to the original checking of the actual value in config or what. 

Any thoughts, @mgeplf / @jdcourcol ?

bluepysnap/simulation.py
joni-herttuainen: You're right. Me neither, didn't even check this. Just assumed this was correct in the first place. I wonder what's the correct/intended behavior here? :thinking:  

Should we even have `time_start` in the `Simulation` class to begin with? :man_shrugging:

bluepysnap/simulation.py
joni-herttuainen: Yeah, `libsonata` returns an object (`libsonata._libsonata.Conditions`) if it was defined and I felt it would be counter-intuitive to return empty dictionary by default.

About breaking changes, this would break the code using it, anyway. We're releasing a new major version, so I am not too worried about small breaking changes. 

I agree, I should mention these in the changelog. I'll update it.

tests/test_simulation.py
joni-herttuainen: That is, again, due to [`libsonata`](https://github.com/BlueBrain/libsonata/blob/master/src/config.cpp#L944): If `node_sets_file` is not defined in the simulation config, `libsonata` fetches it from the circuit config. 

Technically, the correct approach would be to remove it also from the circuit config but this was just simpler. I'll clarify it with a comment.

tests/test_simulation.py
joni-herttuainen: I'll rename the test if need be. We should discuss and decide on the libsonata behavior first, though.

bluepysnap/simulation.py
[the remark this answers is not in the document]
GianlucaFicarelli: ic, it seems ok to me to keep your changes and relying on libsonata using "circuit_config.json" as default, since it's mentioned in the sonata specs.

tests/test_simulation.py
GianlucaFicarelli: Yeah, no need to add `test_no_network_config` as I was thinking in the previous comment. I would just rename the name of the test.

bluepysnap/simulation.py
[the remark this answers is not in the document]
joni-herttuainen: So it seems. I'll keep this behavior, then

tests/test_simulation.py
joni-herttuainen: > renamed

bluepysnap/simulation.py
mgeplf: Open a ticket in libsonata if there is an inconsistency, pls; don't forget to look in https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md as well

joni-herttuainen: Okay. According to jdc, simulation should always start at `0` so this is not in the spec.
I'll have this return 0, then.

GianlucaFicarelli (approved):
LGTM, unless it can wait for the release of `libsonata>=0.1.14` to adapt the retrieval of the filename

joni-herttuainen:
Yeah, I'll just add that to my list and fix it when the next version of libsonata is released. There probably wouldn't be any merge conflicts even if I waited but just in case

PR #166, opened 2022-08-30
Add option to only print/return fatal errors in validation

joni-herttuainen:
This is to ignore warnings and INFO messages (not currently used) raised diring the validation.

Also ignores the datatype warning messages raised during schema validation.

### DIFF
--- a/bluepysnap/cli.py
+++ b/bluepysnap/cli.py
@@ -28,7 +28,7 @@ def cli(verbose):
         "edge indices are correct, etc"
     ),
 )
-@click.option("--only-errors", is_flag=True, help=("Only print fatal errors (ignore warnings)"))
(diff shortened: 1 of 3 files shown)


bluepysnap/cli.py
```diff
@@ -28,11 +28,13 @@ def cli(verbose):
         "edge indices are correct, etc"
     ),
 )
-def validate(config_file, skip_slow):
+@click.option("--only-errors", is_flag=True, help=("Only print fatal errors (ignore warnings)"))
```
GianlucaFicarelli, line 31: Minor, redundant parentheses 
```suggestion
@click.option("--only-errors", is_flag=True, help="Only print fatal errors (ignore warnings)")
```

mgeplf (approved):

GianlucaFicarelli (approved):

PR #170, opened 2022-10-26
Fix tests with new libsonata

GianlucaFicarelli:
Opening as draft until a version of libsonata is released and the skipped tests are restored.

### DIFF
--- a/bluepysnap/nodes.py
+++ b/bluepysnap/nodes.py
@@ -20,7 +20,6 @@ import inspect
 from collections.abc import Mapping, Sequence
 from copy import deepcopy
 
-import libsonata
 import numpy as np
 import pandas as pd
 from cached_property import cached_property
--- a/bluepysnap/nodes.py
+++ b/bluepysnap/nodes.py
@@ -575,11 +575,6 @@ class NodePopulation:
         for node_conf in self._circuit.config["networks"]["nodes"]:
             if self.name in node_conf["populations"]:
                 return node_conf["nodes_file"]
-        for node_conf in self._circuit.config["networks"]["nodes"]:
-            h5_filepath = node_conf["nodes_file"]
-            storage = libsonata.NodeStorage(h5_filepath)
-            if self.name in storage.population_names:  # pylint: disable=unsupported-membership-test
-                return h5_filepath
         raise BluepySnapError(f"h5_filepath not found for population '{self.name}'")
 
 
--- a/tests/data/circuit_config.json
+++ b/tests/data/circuit_config.json
@@ -16,6 +16,9 @@
         "populations": {
             "default": {
                 "type": "biophysical"
+            },
+            "default2": {
+                "type": "biophysical"
             }
         }
       }
@@ -25,6 +28,9 @@
         "edges_file": "$NETWORK_DIR/edges.h5",
         "populations": {
             "default":{
+                "type": "chemical"
+            },
+            "default2":{
                 "type": "chemical"
             }
         }
--- a/tests/test_circuit.py
+++ b/tests/test_circuit.py
@@ -1,6 +1,5 @@
 import json
 
-import libsonata
 import pytest
 from libsonata import SonataError
 
--- a/tests/test_circuit.py
+++ b/tests/test_circuit.py
@@ -1,5 +1,6 @@
 import json
 
+import libsonata
 import pytest
 from libsonata import SonataError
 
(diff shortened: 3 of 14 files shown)


tests/utils.py
```diff
@@ -78,17 +85,18 @@ def create_node_population(filepath, pop_name, circuit=None, node_sets=None, pop
...
         node_sets: (Mock/NodeSets): either a real node_sets or a mocked node_sets.
+        pop_type (str): optional population type.
     Returns:
         NodePopulation: return a node population.
     """
     node_pop_config = {
         "nodes_file": filepath,
         "node_types_file": None,
-        "populations": {},
+        "populations": {pop_name: {}},
```
edasubert, line 95: is the `pop_name` key now mandatory?

tests/data/circuit_config.json
```diff
@@ -16,6 +16,9 @@
         "populations": {
             "default": {
                 "type": "biophysical"
+            },
+            "default2": {
```
edasubert, line 20: `default2` does not really make sense to me as a name, how does one use two defaults? 🙂

tests/test_circuit_ids.py
```diff
@@ -19,7 +19,7 @@ def _multi_index():
 
 
 class TestCircuitNodeId:
-    def setup(self):
+    def setup_method(self):
```
edasubert, line 22: why is the name changed?

edasubert (commented):
I don't have enough context to review the actual removed code, but at least I have several annoying questions/comments 😊

tests/utils.py
GianlucaFicarelli: Population(s) are now required in the sonata extension documented at https://sonata-extension.readthedocs.io/en/latest/sonata_config.html :
> The populations dictionary should contain only the node populations that are part of the circuit. It must contain at least one population, but it doesn’t need to contain all the populations present in the nodes_file if there are more. It cannot be empty because it would mean that it’s incomplete (ie: a population that should have been added, wasn’t), or that the nodes_file shouldn’t be added to the circuit.

tests/data/circuit_config.json
GianlucaFicarelli: Yeah, `default` and `default2` aren't meaningful names, and they don't follow the new naming convention described at https://sonata-extension.readthedocs.io/en/latest/sonata_population.html.

However, this is only for test so the naming is less important.

The names `default` and `default2` are the population names present in the h5 files already used for testing.
You can see their content opening the files with h5py, or with the cli for example with:
```
h5ls -r tests/data/nodes.h5
h5ls -r tests/data/edges.h5
```

tests/test_circuit_ids.py
GianlucaFicarelli: `setup` was specific to nosetests and pytest 7.2.0 raises now a warning if it is used:

```
PytestRemovedIn8Warning: Support for nose tests is deprecated and will be removed in a future release.
  tests/test_morph.py::TestMorphHelper::test_get_filepath is using nose-specific method: `setup(self)`
  To remove this warning, rename it to `setup_method(self)`
  See docs: https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose
    fixture_result = next(generator)
```

bluepysnap/nodes.py
```diff
@@ -575,11 +574,6 @@ def h5_filepath(self):
         for node_conf in self._circuit.config["networks"]["nodes"]:
             if self.name in node_conf["populations"]:
                 return node_conf["nodes_file"]
-        for node_conf in self._circuit.config["networks"]["nodes"]:
-            h5_filepath = node_conf["nodes_file"]
-            storage = libsonata.NodeStorage(h5_filepath)
-            if self.name in storage.population_names:  # pylint: disable=unsupported-membership-test
-                return h5_filepath
```
joni-herttuainen, line 582: Yass!! We'll get rid of this.

tests/test_circuit.py
```diff
@@ -29,6 +29,7 @@ def test_all():
     )
 
 
+@skip_if_libsonata_0_1_16
```
joni-herttuainen, line 32: Neat. I wonder, however, if, starting from the release, we would want to require `libsonata>=0.1.16` and remove these tests altogether. 🤔

joni-herttuainen (approved):
Thanks for fixing the tests.

I'll approve but I feel we want to set a lower limit on the libsonata version before the release. I can test and see to it in another PR.

mgeplf (approved):

mgeplf (approved):
If there are missing tests in libsonata, let me know.

PR #172, opened 2022-10-28
Use libsonata for getting the frame report file name

joni-herttuainen:
* require libsonata>=0.1.16

Since tests are fixed in #170 , this should be merged after that one.


edasubert (approved):
it looks good to me 👍 assuming the tests are indeed fixed in the other branch

I would probably just merge https://github.com/BlueBrain/snap/pull/170 in this branch so the end result in `main` is clear from the MR, but it is probably also ok to just wait

PR #179, opened 2023-02-01
Access population config from the populations

joni-herttuainen:
Added back the ability to access population config from the populations.
* also has the components
* also has nodes/edges file path to avoid having to loop for it in `nodes.py` / `edges.py`

### DIFF
--- a/bluepysnap/circuit.py
+++ b/bluepysnap/circuit.py
@@ -49,6 +49,14 @@ class Circuit:
         """Network config dictionary."""
         return self._config.to_dict()
 
+    def get_node_population_config(self, name):
+        """Get node population configuration."""
+        return self._config.populations["node"].get(name, {})
+
+    def get_edge_population_config(self, name):
+        """Get edge population configuration."""
+        return self._config.populations["edge"].get(name, {})
+
     @cached_property
     def node_sets(self):
         """Returns the NodeSets object bound to the circuit."""
--- a/bluepysnap/circuit.py
+++ b/bluepysnap/circuit.py
@@ -21,6 +21,7 @@ from cached_property import cached_property
 
 from bluepysnap.config import CircuitConfig
 from bluepysnap.edges import Edges
+from bluepysnap.exceptions import BluepySnapError
 from bluepysnap.node_sets import NodeSets
 from bluepysnap.nodes import Nodes
 
@@ -51,11 +52,17 @@ class Circuit:
 
     def get_node_population_config(self, name):
         """Get node population configuration."""
-        return self._config.populations["node"].get(name, {})
+        try:
+            return self._config.populations["node"][name]
+        except KeyError as e:
+            raise BluepySnapError(f"Population config not found for node population: {name}") from e
 
     def get_edge_population_config(self, name):
         """Get edge population configuration."""
-        return self._config.populations["edge"].get(name, {})
+        try:
+            return self._config.populations["edge"][name]
+        except KeyError as e:
+            raise BluepySnapError(f"Population config not found for edge population: {name}") from e
 
     @cached_property
     def node_sets(self):
--- a/bluepysnap/circuit.py
+++ b/bluepysnap/circuit.py
@@ -53,14 +53,14 @@ class Circuit:
     def get_node_population_config(self, name):
         """Get node population configuration."""
         try:
-            return self._config.populations["node"][name]
+            return self._config.node_populations[name]
         except KeyError as e:
             raise BluepySnapError(f"Population config not found for node population: {name}") from e
 
     def get_edge_population_config(self, name):
         """Get edge population configuration."""
         try:
-            return self._config.populations["edge"][name]
+            return self._config.edge_populations[name]
         except KeyError as e:
             raise BluepySnapError(f"Population config not found for edge population: {name}") from e
 
--- a/bluepysnap/config.py
+++ b/bluepysnap/config.py
@@ -187,8 +187,9 @@ class CircuitConfig(Config):
         # components: null in the config
         config = self.to_dict()
         networks = config.get("networks") or {}
+        components = config.get("components") or {}
 
-        def resolve_network_populations(element_type, components):
+        def resolve_network_populations(element_type):
             populations = {}
             type_file_key = f"{element_type}_file"
             element_dicts = networks.get(element_type) or []
@@ -206,10 +207,9 @@ class CircuitConfig(Config):
 
             return populations
 
-        # 'components' is not extended to edges as it currently has no keys for edges.
         self._populations = {
-            "node": resolve_network_populations("nodes", config.get("components") or {}),
-            "edge": resolve_network_populations("edges", {}),
+            "node": resolve_network_populations("nodes"),
+            "edge": resolve_network_populations("edges"),
         }
 
 
--- a/bluepysnap/config.py
+++ b/bluepysnap/config.py
@@ -175,9 +175,15 @@ class CircuitConfig(Config):
         return cls(config_path, libsonata.CircuitConfig)
 
     @property
-    def populations(self):
-        """Access population configs."""
-        return self._populations
+    def node_populations(self):
+        """Access node population configs."""
+        return self._populations['nodes']
+
+    @property
+    def edge_populations(self):
+        """Access edge population configs."""
+        return self._populations['edges']
+
 
     # NOTE: These are parsed here to keep Parser returning the exact parsed configuration.
     def _resolve_population_configs(self):
@@ -205,8 +211,8 @@ class CircuitConfig(Config):
             return populations
 
         return {
-            "node": resolve_network_populations("nodes"),
-            "edge": resolve_network_populations("edges"),
+            "nodes": resolve_network_populations("nodes"),
+            "edges": resolve_network_populations("edges"),
         }
 
 
--- a/bluepysnap/config.py
+++ b/bluepysnap/config.py
@@ -167,7 +167,7 @@ class CircuitConfig(Config):
     def __init__(self, *args):
         """Initializes circuit config."""
         super().__init__(*args)
-        self._populations = self._resolve_population_configs()
+        self._populations = self._resolve_population_configs(self.to_dict())
 
     @classmethod
     def from_config(cls, config_path):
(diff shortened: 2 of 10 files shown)


bluepysnap/nodes.py
```diff
@@ -571,10 +580,10 @@ def models(self):
     @cached_property
```
GianlucaFicarelli, line 580: Same comment as for edges

bluepysnap/nodes.py
```diff
@@ -571,10 +580,10 @@ def models(self):
...
     def h5_filepath(self):
         """Get the H5 nodes file associated with population."""
-        for node_conf in self._circuit.config["networks"]["nodes"]:
-            if self.name in node_conf["populations"]:
-                return node_conf["nodes_file"]
-        raise BluepySnapError(f"h5_filepath not found for population '{self.name}'")
+        try:
+            return self.population_config["nodes_file"]
+        except KeyError as e:
+            raise BluepySnapError(f"h5_filepath not found for population '{self.name}'") from e
```
GianlucaFicarelli, line 586: Same comment as for edges

bluepysnap/config.py
```diff
@@ -163,11 +164,54 @@ def to_dict(self):
 class CircuitConfig(Config):
     """Handle CircuitConfig."""
 
+    def __init__(self, *args):
+        """Initializes circuit config."""
+        super().__init__(*args)
+        self._populations = None
+        self._resolve_population_configs()
```
GianlucaFicarelli, line 171: Minor, if `_resolve_population_configs` returns a dict, it could be replaced with:
```python
self._populations = self._resolve_population_configs()
```
that would make clearer that `self._populations` is never None.

bluepysnap/circuit.py
```diff
@@ -49,6 +49,14 @@ def config(self):
         """Network config dictionary."""
         return self._config.to_dict()
 
+    def get_node_population_config(self, name):
+        """Get node population configuration."""
+        return self._config.populations["node"].get(name, {})
+
+    def get_edge_population_config(self, name):
+        """Get edge population configuration."""
+        return self._config.populations["edge"].get(name, {})
```
GianlucaFicarelli, line 58: I think these methods should raise a BluepySnapError if the population is not found.

bluepysnap/config.py
```diff
@@ -163,11 +164,54 @@ def to_dict(self):
...
+
+            for elem_dict in element_dicts:
+                elem_pops = elem_dict.get("populations") or {}
+
+                for pop, cfg in elem_pops.items():
+                    populations[pop] = copy.deepcopy(components)
+                    populations[pop].update(cfg or {})
+
+                    # add h5 filepath for simpler access
+                    if type_file_key in elem_dict:
```
GianlucaFicarelli, line 205: Since `nodes_file` and `edges_file` are mandatory, I think we can avoid the `if`.

bluepysnap/config.py
```diff
@@ -163,11 +164,54 @@ def to_dict(self):
...
+        # All the "or"s exist to have empty dicts lists instead of None's to overcome issues like:
+        # components: null in the config
+        config = self.to_dict()
+        networks = config.get("networks") or {}
+        components = config.get("components") or {}
+
+        def resolve_network_populations(element_type):
+            populations = {}
+            type_file_key = f"{element_type}_file"
+            element_dicts = networks.get(element_type) or []
```
GianlucaFicarelli, line 195: `nodes` and `edges` keys should be mandatory, so `get` and `or` shouldn't be needed.

bluepysnap/config.py
```diff
@@ -163,11 +164,54 @@ def to_dict(self):
...
+
+            for elem_dict in element_dicts:
+                elem_pops = elem_dict.get("populations") or {}
```
GianlucaFicarelli, line 198: `populations` is mandatory, so the get shouldn't be needed

bluepysnap/config.py
```diff
@@ -163,11 +164,54 @@ def to_dict(self):
...
+        """Access population configs."""
+        return self._populations
+
+    # NOTE: These are parsed here to keep Parser returning the exact parsed configuration.
+    def _resolve_population_configs(self):
+        """Resolves population configs for the node and edge populations."""
+        # All the "or"s exist to have empty dicts lists instead of None's to overcome issues like:
+        # components: null in the config
+        config = self.to_dict()
+        networks = config.get("networks") or {}
```
GianlucaFicarelli, line 189: `networks` is mandatory, so `get` and `or` shouldn't be needed

bluepysnap/config.py
```diff
@@ -163,11 +164,54 @@ def to_dict(self):
...
+            populations = {}
+            type_file_key = f"{element_type}_file"
+            element_dicts = networks.get(element_type) or []
+
+            for elem_dict in element_dicts:
+                elem_pops = elem_dict.get("populations") or {}
+
+                for pop, cfg in elem_pops.items():
+                    populations[pop] = copy.deepcopy(components)
+                    populations[pop].update(cfg or {})
```
GianlucaFicarelli, line 202: According to the specs I don't think that `cfg` can be None, so the `or` shouldn't be needed.

bluepysnap/config.py
joni-herttuainen: Yeah, good call! I'll change it. I think I just tunnel-visioned myself out here.

bluepysnap/config.py
joni-herttuainen: Yeah, so, did I. However, the spec does not forbid a config with something like:
```json
...
"populations": {
    "default": null
}
```
In which case this would happen and, as a matter of fact, that is what happens in tests.

joni-herttuainen (commented):
Regarding all the missing field checks (`or {}`/`[]`), if the config is `partial`, it may be missing:
* `networks`
* `edges`
* `nodes`
* `edges_file`
* `nodes_file`
* `populations`

Do we want/is there a need to support partial configs with SNAP?

joni-herttuainen:
We would likely be supporting partial configs in SNAP, so I'd rather leave the "safety checks" for now. If at some point they're not needed, they are easy to remove.

bluepysnap/config.py
GianlucaFicarelli: About dumping to json, it's weird 🤔 because
```python
json.dumps({"populations": {"default": {}}})
```
gives
```
'{"populations": {"default": {}}}'
```
so there should be something in the middle doing some transformation.

However, keeping the `or` fallback doesn't hurt 👍

GianlucaFicarelli:
> We would likely be supporting partial configs in SNAP, so I'd rather leave the "safety checks" for now. If at some point they're not needed, they are easy to remove.

Ok it sounds good to me.

bluepysnap/config.py
joni-herttuainen: I rechecked the test. I misremembered. It wasn't the `json.dumps` that would cause the issue, it would be `libsonata.CircuitConfig.expanded_json` (or something that creates the output) that causes the coercion from `{}` to `null`.

bluepysnap/edges.py
```diff
@@ -566,10 +575,10 @@ def iter_connections(
     @cached_property
```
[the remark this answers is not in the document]
joni-herttuainen: -> done

joni-herttuainen:
Since libsonata does not fully support partial configs, or at least does not yet have a release that does, I created a ticket [NSETM-2114](https://bbpteam.epfl.ch/project/issues/browse/NSETM-2114) for this and just keep assuming all configs are "complete" for now.

bluepysnap/config.py
```diff
@@ -163,11 +164,51 @@ def to_dict(self):
...
+
     @classmethod
     def from_config(cls, config_path):
         """Instantiate the config class from circuit configuration."""
         return cls(config_path, libsonata.CircuitConfig)
 
+    @property
+    def populations(self):
+        """Access population configs."""
+        return self._populations
```
edasubert, line 180: why not just access `populations` directly? (make the attribute name without an underscore)

bluepysnap/edges.py
```diff
@@ -114,6 +114,14 @@ def _dynamics_params_names(self):
     def _topology_property_names(self):
         return {Edge.SOURCE_NODE_ID, Edge.TARGET_NODE_ID}
 
+    @property
+    def population_config(self):
+        """Access the configuration for the population.
+
+        This configuration is extended the filepath to the h5 file containing the population.
```
edasubert, line 121: I do not understand this line of the comment :slightly_smiling_face:

tox.ini
```diff
@@ -63,7 +63,8 @@ max-line-length = 100
 [pydocstyle]
 # ignore the following
 #   - D413: no blank line after last section
-add-ignore = D413
+#   - D202: no blank line after docstring (handled by black, causes an issue if first line is '#pylint:disable')
```
edasubert, line 66: it can be solved by putting the disable above the function definition, but I am fine with this ignore :slightly_smiling_face:

bluepysnap/edges.py
joni-herttuainen: Me neither. I clarified it a good bit.

bluepysnap/config.py
joni-herttuainen: I feel this is a more pythonic way and in line with the rest of the code. 

Also, I like keeping things safe, so that external classes/methods/functions don't directly access the properties. Serving it as a property removes the possibility to mess it up with something like
`config.populations = <whatever>`

bluepysnap/config.py
```diff
@@ -163,11 +164,51 @@ def to_dict(self):
...
+
+                    # add h5 filepath for simpler access
+                    if type_file_key in elem_dict:
+                        populations[pop][type_file_key] = elem_dict[type_file_key]
+
+            return populations
+
+        return {
+            "node": resolve_network_populations("nodes"),
+            "edge": resolve_network_populations("edges"),
```
GianlucaFicarelli, line 209: Minor, keeping the plural names "nodes" and "edges" seems more natural to me, but it can be subjective.

GianlucaFicarelli (commented):
LGTM

tox.ini
GianlucaFicarelli: Moving the disable above the function won't have the same effect, because it would affect the full block and not only the single method (i.e. the disable affects all the methods defined after the disable).

bluepysnap/config.py
joni-herttuainen: Good point, I'll add a few tests to make sure that the components and population config is done in the correct order etc., so might as well change this to be uniform with everything else.

edasubert (approved):
:+1:

bluepysnap/config.py
```diff
@@ -163,11 +164,57 @@ def to_dict(self):
...
+    def node_populations(self):
+        """Access node population configs."""
+        return self._populations["nodes"]
+
+    @property
+    def edge_populations(self):
+        """Access edge population configs."""
+        return self._populations["edges"]
+
+    # NOTE: These are parsed here to keep Parser returning the exact parsed configuration.
```
GianlucaFicarelli, line 187: I don't understand this comment 🤔 
Can it be clarified and moved into the relevant method, or removed?

joni-herttuainen: Yeah, I removed the comment as it was more confusing than informative.

Short story time: I was initially thinking of parsing the population configs in the `class Parser` and returning them as extra parts in the config, but then realized it may or may not break stuff and/or make stuff more confusing so I added a specific method to `class CircuitConfig`. Hence the comment.

GianlucaFicarelli (approved):

PR #180, opened 2023-02-03
HOTFIX: omit dtypes when doing nodes.get or edges.get [NSETM-2112]

joni-herttuainen:
This is a temporary (as if) fix to the "Same property with different dtype" error. 

Need to figure how to fix this while still keeping the efficiency of @GianlucaFicarelli 's optimisations. It is possible to merge categories of catecorigals but it seems it's not as straightforward as we'd hope.

So to have the get functions working. We should go with this for now, release 1.0.1 and add a proper fix when someone has time to work on this.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+Version v1.0.1
+--------------
+
+Bug Fixes
+~~~~~~~~~
+- nodes.get, edges.get were throwing an incorrect "Same property with different dtype" Error
+
 Version v1.0.0
 --------------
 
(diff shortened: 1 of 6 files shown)


edasubert (approved):
:+1:

GianlucaFicarelli:
Unless it's urgent, I'd prefer to wait before applying this hotfix, until it's clarified what should be the desired behaviour when the dtypes don't match, for both `get()` and `property_dtypes` (that would still raise an Exception with these changes).
We should also add a test to cover this scenario.

joni-herttuainen:
@GianlucaFicarelli : Well, we can always tell people to use `circuit.nodes["<node_pop>"].get` for now, if we wish to not roll back to previous behavior.

I am not really familiar with many use cases in which this would pose an issue otherwise. I guess that would be the case if node sets combine multiple h5 files/populations that had different categories / dtypes.

joni-herttuainen:
Fixed in #217

PR #184, opened 2023-02-22
Teach schemas to return datatypes for attributes for the nodes and edges

mgeplf:


bluepysnap/schemas/schemas.py
```diff
@@ -206,3 +208,65 @@ def validate_edges_schema(path, edges_type, virtual):
...
+    Returns:
+        dict: name -> type of column
+    """
+    if virtual:
+        edges_type += "_virtual"
+
+    schema = _parse_schema("edge", edges_type)
+    resolver = jsonschema.validators.RefResolver("", schema)
+
+    schema = schema["$edge_file_defs"]["edges_file_root"]["properties"]["edges"]
```
mgeplf, line 267: not super happy about having to dig so deep into the dictionary - if you have a better idea on how to do this, that would be cool.

bluepysnap/schemas/schemas.py
```diff
@@ -206,3 +208,65 @@ def validate_edges_schema(path, edges_type, virtual):
...
+    del schema["@library"]
+
+    return _resolve_types(resolver, schema), _resolve_types(resolver, dynamics_params)
+
+
+def edges_schema_types(edges_type, virtual):
+    """Get the datatypes of the attribute for nodes.
+
+    Args:
+        nodes_type (str): node type (e.g., "chemical")
```
joni-herttuainen, line 255: minor: `edges_type (str): edge type (e.g., "biophysical")`

bluepysnap/schemas/schemas.py
joni-herttuainen: Yeah, I get it. 

If the use case is just to get a mapping of a node/edge property to a dtype (i.e., we don't actually care about the edge/node type), we could have the properties of `<edge|node>/<population_name>/0` in their own files. Something like `{edge,node}_0_properties.yaml`.

And then just add a reference to those in the `{edge,node}.yaml` files.

That would kill two birds with one stone: 
* make it simpler to do the parsing of the properties here
* make the `{edge,node}.yaml` files more readable

I'm also not super happy of the long list of properties, or rather the fact that we need to do:
```yaml
property_1:
     $ref: dtype_1
property_2:
     $ref: dtype_1
...
```
if we could have something like
```yaml
dtype_1:
  - property_1
  - property_2
 ...
```
it would be cleaner, IMO. Not sure if this is easily doable but I'll give it some thought. For sure it is, if we make some custom parsing before feeding it to `jsonschema` but that is the route I'd rather avoid if possible.

mgeplf: I guess we can do this reorg some other time; I understand better now the structure, and why it is like it is.

I'm going to play w/ this a bit more, to see if it works for my use case.

bluepysnap/schemas/definitions/edge.yaml
```diff
@@ -49,55 +49,92 @@ $edge_file_defs:
               source_node_id:
                 $ref: "#/$edge_file_defs/edge_node_ids"
 
-              target_node_id:
+              targe_node_id:
```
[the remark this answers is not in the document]
mgeplf: tnx!

bluepysnap/schemas/schemas.py
mgeplf: tnx

bluepysnap/schemas/schemas.py
joni-herttuainen: Sure. fine with me

joni-herttuainen (approved):

bluepysnap/schemas/definitions/edge.yaml
```diff
@@ -49,55 +49,92 @@ $edge_file_defs:
...
               edge_type_id:
                 $ref: "#/$typedefs/int64"
 
               "0":
                 type: object
-                patternProperties:
+                properties:
                   # For docs, these would likely need to be changed actual property names.
                   # Best probably combine each to a definition like "#/edge_field_defs/..._xyz"
                   # to keep stuff cleaner as they will be reused anyhow.
```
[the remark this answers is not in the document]
mgeplf: fixed

bluepysnap/schemas/schemas.py
```diff
@@ -193,6 +194,7 @@ def validate_edges_schema(path, edges_type, virtual):
     Args:
         path (str): path to the edges file
         edges_type (str): edge type (e.g., "chemical")
+        virtual(bool): whether this a virtual edge population
```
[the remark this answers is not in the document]
mgeplf: tnx, fixed

PR #190, opened 2023-03-20
add pickling support to most classes with caches

mgeplf:
cuts down on mistakes when using pickling


mgeplf:
@eleftherioszisis can you try this with connectome-manipulator, and see if it solves the pickle problem?

mgeplf:
nice, seems like an improvement, thanks for checking.

edasubert (approved):
really cool :+1:

joni-herttuainen (approved):

PR #192, opened 2023-03-28
Added intersection() method to circuit ids

HDictus:
`np.intersect1d(ids, other_ids)` was a really common tool in the bluepy.v2 days . 
This is intended to replace that operation for bluepysnap.
It gets a CircuitIds object which has only those ids in both the calling object and the passed object.

### DIFF
--- a/bluepysnap/circuit_ids.py
+++ b/bluepysnap/circuit_ids.py
@@ -263,11 +263,7 @@ class CircuitIds(abc.ABC):
             circuit_ids (CircuitIds): The CircuitIds to intersect with.
             inplace (bool): if set to True, do the transformation inplace.
         """
-
-        def _intersection(index):
-            return index.intersection(circuit_ids.index)
-
-        return self._apply(_intersection, inplace)
+        return self._apply(lambda x: x.intersection(circuit_ids.index), inplace)
 
     def to_csv(self, filepath):
         """Save CircuitIds to csv format."""
--- a/bluepysnap/circuit_ids.py
+++ b/bluepysnap/circuit_ids.py
@@ -258,8 +258,7 @@ class CircuitIds(abc.ABC):
 
         Notes:
             This function returns a CircuitIDs which contains only those ids which are in both
-            the this CircuitNodeIDs and the provided CircuitIDs. The index of the resulting
-            CircuitNodeIDs will be sorted.
+            this CircuitIds and the provided CircuitIds. The index of the result will be sorted.
 
         Args:
             ids: The other CircuitNodeIDs
--- a/bluepysnap/circuit_ids.py
+++ b/bluepysnap/circuit_ids.py
@@ -256,18 +256,16 @@ class CircuitIds(abc.ABC):
     def intersection(self, circuit_ids, inplace=False):
         """Take the intersection of this CircuitIds and the input.
 
-        Notes:
-            The index of the result is sorted.
+        The index of the resulting object is sorted if ``inplace=False``.
+        Otherwise, the orginal order of the index is kept.
 
         Args:
             circuit_ids (CircuitIds): The CircuitIds to intersect with.
             inplace (bool): if set to True, do the transformation inplace.
         """
-        ids = set(circuit_ids.tolist())
 
-        def _intersection(ind):
-            res = ids.intersection(set(ind))
-            return self.from_tuples((*res,), sort_index=True).index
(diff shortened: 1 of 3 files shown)


mgeplf:
Can you have a look at this, @joni-herttuainen ?

joni-herttuainen:
I can. I was just thinking: suppose we remove the `edges.get` and `nodes.get` methods, should that not also render the `ids` methods useless and, therefore, the `CircuitIds`?

bluepysnap/circuit_ids.py
```diff
@@ -253,6 +253,26 @@ def unique(self, inplace=False):
         """
         return self._apply(lambda x: x.unique(), inplace)
 
+    def intersection(self, ids, inplace=False):
+        """Take the intersection of this CircuitIDs with another.
+
+        Notes:
+            This function returns a CircuitIDs which contains only those ids which are in both
+            this CircuitIds and the provided CircuitIds. The index of the result will be sorted.
```
joni-herttuainen, line 261: Is there a specific reason why this is sorted by default?

HDictus: would it be better if it did not sort?

joni-herttuainen: I don't know, really. In the very worst scenario, sorting could add quite the overhead but realistically thinking, in most cases the ids are likely sorted or mostly sorted anyway. Let's keep it sorted for now.

joni-herttuainen:
We need to get nodes.get and edges.get working first.

HDictus:
you mean it needs to  be decided whether or not they'll be removed?
What will replace them if so?

joni-herttuainen (approved):
I checked that [intersect1d](https://numpy.org/doc/stable/reference/generated/numpy.intersect1d.html) also returns sorted list so it makes sense your implementation has the behavior. 

I'll take my earlier comment back. We can merge this regardless of what we are going to do with the other issues.

If anything, just add a mention in the changelog.

bluepysnap/circuit_ids.py
```diff
@@ -253,6 +253,26 @@ def unique(self, inplace=False):
...
+            inplace: if True, modify this CircuitNodeIDs
+        """
+        tuples_other = set(ids.tolist())
+
+        def _intersect(ind):
+            tuples_self = set(ind)
+            intersected = tuples_self.intersection(tuples_other)
+            return self.from_tuples(list(intersected), sort_index=True).index
+
+        return self._apply(_intersect, inplace)
```
GianlucaFicarelli, line 274: Pandas provides an `intersection` method that works with both Index and MultiIndex, and it's probably more efficient than converting the index to a set.

```suggestion
        return self._apply(lambda x: x.intersection(ids.index), inplace)
```

About sorting:
if inplace=False, then a new CircuitIds is created (using the default sort_index=True).
if inplace=True, then the existing CircuitIds is used (and we don't know if it's already sorted or not, unless we store the information when the object is initialized, or calculate it)

It would be simpler if we remove the option to create CircuitIds with sort_index=False (i.e. CircuitIds is always sorted), but I wonder if there is any use case that justifies keeping this option.

tests/test_circuit_ids.py
```diff
@@ -209,6 +209,16 @@ def test_append(self):
         )
         assert test_obj == expected
 
+    def test_intersection(self):
+        first_ids = self.ids_cls.from_tuples([("a", 1), ("a", 2), ("a", 1), ("a", 2), ("b", 1)])
+        to_intersect = self.ids_cls.from_tuples([("b", 1), ("a", 3), ("a", 2)])
+        expected = self.ids_cls.from_tuples([("a", 2), ("b", 1)])
+        print(first_ids, first_ids.intersection(to_intersect))
```
GianlucaFicarelli, line 216: The print statement is probably a leftover and it should be removed.

joni-herttuainen: agreed -> removing

bluepysnap/circuit_ids.py
joni-herttuainen: Sure, I'll take a look and run a few small tests

joni-herttuainen: Yeah, pandas was like 30-40% faster, so implemented that, tthanks!

GianlucaFicarelli (approved):
LGTM

PR #194, opened 2023-04-13
NSETM-2114 partial circuit config

edasubert:


tests/test_config/test_partial_config.py
```diff
@@ -0,0 +1,118 @@
...
+            "components": {
+                "morphologies_dir": "some/morph/dir",
+            },
+            "networks": {
+                "edges": [],
+            },
+        },
+    ],
+)
+def test_partial_circuit_config_empty(config):
```
edasubert, line 46: If nodes/edges listing is missing or it is empty, empty lists or errors are returned (I have added one). Unless we want to disable Circuit running empty entirely, I think it can stays as is.

tests/test_config/test_partial_config.py
```diff
@@ -0,0 +1,118 @@
...
+        circuit.edges.ids()
+
+
+@pytest.mark.parametrize(
+    "nodes_update, edges_update",
+    [
+        ({"populations": {"default": {}}}, {"populations": {"default": {}}}),
+    ],
+)
+def test_partial_circuit_config_partial(nodes_update, edges_update):
```
edasubert, line 79: I did not figure out other ways to make the configuration partial, ideas are welcome

It seems that the `type` is guessed and everything is working.

tests/test_config/test_partial_config.py
```diff
@@ -0,0 +1,118 @@
...
+            assert circuit.get_edge_population_config("default")
+            assert circuit.edges
+            assert circuit.edges["default"].type == "chemical"
+            assert circuit.edges.population_names == ["default"]
+            assert list(circuit.edges.values())
+
+            assert [item.id for item in circuit.edges.ids()] == [0, 1, 2, 3]
+
+
+def test_partial_circuit_config_full():
```
edasubert, line 108: this is just a full config with the partial flag, not sure this test makes any sense

edasubert:
@mgeplf @joni-herttuainen I created few tests for the partial circuit config; feedback on what more to include or what is perhaps unnecessary is welcome :slightly_smiling_face:

joni-herttuainen:
I think the easiest would be to make a list of things that may be missing in a partial circuit and see that those cases are covered. I guess anything that is mandatory can be missing from a partial circuit but what are the realistical use cases? E.g., in [NSETM-2003](https://bbpteam.epfl.ch/project/issues/browse/NSETM-2003), it is mentioned that:
> things like emodels may not exist for the biophysical populations

So that, I think, should be covered. I don't think it's realistic use case to have biophysical node population + files without there being morphologies so I don't think missing morph files need to be covered.

Furthermore, I think we should show a clear warning when a partial circuit is loaded and mention about the consequences. I think it would be worth it to do this also when trying to validate a partial circuit.

bluepysnap/circuit_validation.py
```diff
@@ -619,6 +623,14 @@ def validate(config_file, skip_slow, only_errors=False, print_errors=True):
     if "networks" in config:
         errors += validate_networks(config, skip_slow)
 
+    if _check_partial_circuit_config(config):
```
edasubert, line 626: I was thinking of converting FATAL errors into WARNINGS for partial configs, but it might be unnecessarily complex

tests/test_config/test_partial_config.py
```diff
@@ -0,0 +1,98 @@
...
+                {
+                    # missing `edges_file` rises libsonata exception
+                    "edges_file": str(TEST_DATA_DIR / "edges.h5"),
+                    "populations": {
+                        "default": {},
+                    },
+                }
+            ],
+        },
+    }
```
edasubert, line 33: I believe this is the smallest config that does not produce errors in libsonata

I have tested everything I found on `NodePopulation` and `EdgePopulation` and nothing fails and everything returns reasonable values

tests/test_edges/test_edge_population.py
```diff
@@ -655,7 +655,7 @@ def test_pickle(self, tmp_path):
         with open(pickle_path, "rb") as fd:
             edge_population = pickle.load(fd)
 
-        assert pickle_path.stat().st_size < 250
+        assert pickle_path.stat().st_size < 260
```
edasubert, line 658: adding the config status increased the size

edasubert:
tests are failing because of `from libsonata import CircuitConfigStatus` which is going to be available after new release, for the time being I could put in `from libsonata._libsonata import CircuitConfigStatus` instead, but I would rather not

mgeplf (approved):

Issue #198, opened 2023-05-02
circ_extr.nodes.property_dtypes throws BluepySnapError: Same property with different dtype. model_template: category!= category


markovg:
Hi,
I got the Zenodo SSCx anatomy Circuit object to create, 
See: https://github.com/BlueBrain/snap/issues/197

but now I'm trying to run the first query command in the example notebook, and 
'circ_extr.nodes.property_dtypes' throws a BluepySnapError: Same property with different dtype. model_template: category!= category
'circ_extr.nodes.property_names'  seems to work fine.

markovg:
Screenshot:
![image](https://user-images.githubusercontent.com/427643/235688105-90d7c2c2-b715-44ff-90d7-d951d454821e.png)

markovg:
This is blocking this line from running:
![image](https://user-images.githubusercontent.com/427643/235688390-ef8214f5-381a-4b0d-93b9-bc018f6eb2f6.png)

markovg:
Apparently it calls "property_dtypes" under the hood

markovg:
Here a deeper look at the issue:
![image](https://user-images.githubusercontent.com/427643/235717716-110508e1-2cc2-47d1-9cf0-ae4b778eda7c.png)

markovg:
It looks to be a mismatch between the CategoricalDType of "S1nonbarrel_neurons" population and the other pops with respect to the "model_template" element.  They are all CategoricalDTypes, but with different categories (empty, or non-empty).  Should it just ignore the contents, or merge, or what? How is this used down stream?  In any case, it looks like _update would need additional logic to handle CategoricalDType?

joni-herttuainen:
Hi @markovg,

Thank you for reporting the issue.  Unfortunately, this is a known issue that we are currently trying to find the best solution to (see: #186).

For now, if you know which population you are dealing with, I suggest you specify that when getting the nodes:
```python
data = circuit.nodes['<population_name>'].get(...)
```
or, if you need to combine data from multiple populations, iterate over the populations:
```python
data = [node_pop.get(...) for node_pop in circuit.nodes.values()] #list of dataframes, needs to be merged
```

markovg:
The issue is arising for the example notebook for the SSCx circuit on Zenodo:
https://zenodo.org/record/6906785/files/Circuit%20exploration%20examples.ipynb?download=1

Pretty much the first line:
`lst_properties = ['layer', 'region', 'synapse_class', 'population',
[...]

PR #199, opened 2023-05-02
 Workaround to issue https://github.com/BlueBrain/snap/issues/198

markovg:
As discussed in the issue, I replace the DType conflict error with a warning.


joni-herttuainen:
Fixed in #217

PR #200, opened 2023-05-03
Index of the output of NodePopulation.get now has a name.

MWolfR:
The naming of the index was introduced for consistency with the output of Nodes.get(). The later returns a DataFrame with a MultiIndex with named levels.

Suppose `circ` is a Circuit.
`nrn1 = circ.nodes.get()`
`nrn2 = circ.nodes["my_population"].get()`

Before, if I do:
`nrn1.reset_index()`
the node ids of the neurons are in a column called "node_ids". While:
`nrn2.reset_index()`
the node ids are in a column called "index".

This is inconsistent in my opinion. I fixed it by naming the index explicitly "node_ids".

Ideally, that specific string would be imported from bluepysnap/circuit_ids.py. To do that I would have to edit that file slightly, as currently the string is implicitly defined in line 283 of that file.

### DIFF
--- a/bluepysnap/nodes/node_population.py
+++ b/bluepysnap/nodes/node_population.py
@@ -100,7 +100,7 @@ class NodePopulation:
         categoricals = nodes.enumeration_names
 
         _all = nodes.select_all()
-        result = pd.DataFrame(index=pd.RangeIndex(_all.flat_size))
+        result = pd.DataFrame(index=pd.RangeIndex(_all.flat_size, name="node_ids"))
 
         for attr in sorted(nodes.attribute_names):
             if attr in categoricals:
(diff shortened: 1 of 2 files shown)


joni-herttuainen:
Hi, thanks for the PR. I see no reason why not add this so let's just go with it.

> Ideally, that specific string would be imported from bluepysnap/circuit_ids.py. To do that I would have to edit that file slightly, as currently the string is implicitly defined in line 283 of that file.

I agree, it would be best if it was defined as a variable to make sure it stays consistent.

MWolfR:
I amended the change such that it takes the name of the index from bluepysnap/circuit_ids.py
Let me know if the way it's implemented makes sense to you, @joni-herttuainen .

joni-herttuainen (approved):
Yeah, looks good to me

PR #208, opened 2023-05-11
Support lazy loading of nodes attributes

GianlucaFicarelli:
~Need more tests to verify the performance.~
See tests in NSETM-2184

### DIFF
--- a/bluepysnap/nodes/node_population.py
+++ b/bluepysnap/nodes/node_population.py
@@ -175,17 +175,16 @@ class NodePopulation:
             nodes.attribute_names,
             utils.add_dynamic_prefix(nodes.dynamics_attribute_names),
         ]
-        # Reverse the iterator to be able to insert columns without invalidating the order.
-        for loc, name in reversed(
-            list(
-                self._filter_properties(
-                    attrs_list=attrs_list,
-                    existing_columns=result.columns,
-                    properties_set=properties_set,
-                )
+        # insert columns at the correct position
+        for n, (loc, name) in enumerate(
+            self._filter_properties(
+                attrs_list=attrs_list,
+                existing_columns=result.columns,
+                properties_set=properties_set,
             )
         ):
-            result.insert(loc, name, self._get_values_from_sonata(name, selection=selection))
+            values = self._get_values_from_sonata(name, selection=selection)
+            result.insert(n + loc, name, values)
         return result
 
     @property
--- a/bluepysnap/nodes/node_population.py
+++ b/bluepysnap/nodes/node_population.py
@@ -173,7 +173,7 @@ class NodePopulation:
             # Select the ids from the cached dataframe.
             # The original dataframe won't be updated in this case.
             result = result.loc[node_ids]
-        cached_columns = result.columns.intersection(properties_set)
+        cached_columns = properties_set.intersection(result.columns)
         if len(cached_columns) < len(properties_set):
             # some requested properties miss from the cache
             nodes = self._population
@@ -456,7 +456,7 @@ class NodePopulation:
         else:
             return np.unique(result)
 
-    def get(self, group=None, properties=None, populate_cache=False):
+    def get(self, group=None, properties=None, populate_cache=True):
         """Node properties as a pandas Series or DataFrame.
 
         Args:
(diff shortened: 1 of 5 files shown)


bluepysnap/nodes/node_population.py
```diff
@@ -94,28 +95,102 @@ def _node_sets(self):
...
+            self._check_properties(properties_set)
+        if node_ids is not None:
+            # Select the ids from the cached dataframe.
+            # The original dataframe won't be updated in this case.
+            result = result.loc[node_ids]
+        cached_columns = properties_set.intersection(result.columns)
+        if len(cached_columns) < len(properties_set):
+            # some requested properties miss from the cache
+            nodes = self._population
+            attrs_list = [
```
mgeplf, line 180: since it's important that this is always stable (since `_iter_selected_properties` depends on it), should it be a cached method, or embedded in `_iter_selected_properties`?

bluepysnap/nodes/node_population.py
```diff
@@ -94,28 +95,102 @@ def _node_sets(self):
...
+            if attr not in desired:
+                continue
+            yield idx, attr
+
+    def _get_data(self, properties=None, node_ids=None):
+        """Collect data for the node population as a pandas.DataFrame.
+
+        Return a DataFrame with node_ids as index, loading the requested properties if needed.
+
+        The returned DataFrame isn't filtered by columns, so it may contain more properties than
```
mgeplf, line 157: perhaps explain why it's not filtered by column? I assume it's for efficiency, since (perhaps?) some calls don't need a copy of the data? maybe say which ones?

bluepysnap/nodes/node_population.py
```diff
@@ -94,28 +95,102 @@ def _node_sets(self):
...
-        for attr in sorted(utils.add_dynamic_prefix(nodes.dynamics_attribute_names)):
-            result[attr] = nodes.get_dynamics_attribute(attr.split(DYNAMICS_PREFIX)[1], _all)
+    def _cache(self):
+        """Cached DataFrame of nodes, to be accessed through _get_data()."""
+        return pd.DataFrame(index=pd.RangeIndex(self.size, name="node_ids"))
+
+    def _get_libsonata_selection(self, node_ids):
+        """Return a libsonata Selection from the given node_ids."""
+        if node_ids is None:
+            return libsonata.Selection([(0, self.size)])
```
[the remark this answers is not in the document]
GianlucaFicarelli: Using directly `libsonata.Selection([(0, self.size)])` is 100x faster because `self._population` isn't cached, and it's still 10x faster even when `self._population` is cached:

```
In [1]: import bluepysnap
   ...: import libsonata
   ...: 
   ...: path = "/gpfs/bbp.cscs.ch/project/proj30/tickets/BLPY-289_improve_get/circuit/circuit_config.json"
   ...: c = bluepysnap.Circuit(path)
   ...: nodes = c.nodes["root__neurons"]
   ...: 
   ...: print(nodes.size)
   ...: _population = nodes._population
   ...: 
   ...: %timeit libsonata.Selection([(0, nodes.size)])
   ...: %timeit _population.select_all()
   ...: %timeit nodes._population.select_all()
   ...: 
74124987
1.25 µs ± 13.8 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
14.5 µs ± 22 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
136 µs ± 81.7 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
However, the times are under 1 ms, so the performance shouldn't be an issue, unless some loop is involved (as in the tests in https://github.com/BlueBrain/snap/pull/211)
Do you still prefer using `self._population.select_all()` in this case?

bluepysnap/nodes/node_population.py
GianlucaFicarelli: Yes it's for efficiency, I added more details

bluepysnap/nodes/node_population.py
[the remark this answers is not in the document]
mgeplf: ok, good to know; keep the current one then :)

mgeplf (approved):

Issue #195, opened 2023-04-20
Performance issues when accessing `positions` for large node populations


matz-e:
While investigating a performance bottleneck transforming morphologies of a circuit, we came across the following part of the code:

https://github.com/BlueBrain/snap/blob/839c910eac56e21f2e09850fba47d4febadcbbde/bluepysnap/morph.py#L97-L98

Accessing the `positions` function in this snipped takes a very long time. Constructing a minimal reproducer:
```
import bluepysnap as snappy
Node = snappy.sonata_constants.Node
import random
import sys
import time

pth = "/gpfs/bbp.cscs.ch/project/proj134/scratch/zisis/connectome-manipulator/circuit/circuit_config.json"
crc = snappy.circuit.Circuit(pth)
nodes = crc.nodes["root__neurons"]
node_ids = [random.randrange(nodes.size) for _ in range(100)]

if sys.argv[1:]:
    ts = time.time()
    for nid in node_ids:
        nodes.get(nid, Node.X)
        nodes.get(nid, Node.Y)
        nodes.get(nid, Node.Z)
    print("single", time.time() - ts)
else:
    ts = time.time()
    for nid in node_ids:
        nodes.positions(nid)
    print("joined", time.time() - ts)
```
And running it:
```
❯ python test.py
joined 152.9171633720398

❯ python test.py 123
single 9.43373441696167
```

Looking through the code for `positions`, it seems that we first construct a new dataframe with the required properties and selecting afterwards? Would be great if this could be switched around.

mgeplf:
Thanks for the nice repro, I will have a look.

mgeplf:
@matz-e can you try https://github.com/BlueBrain/snap/pull/196/ and see if it helps?

matz-e:
I ran this with our test case, and we get some nice improvement all over the place with a total of ~30% runtime improvement! (even better than my "fix" above)

PR #211, opened 2023-05-22
Cache properties accessing libsonata methods

GianlucaFicarelli:
If nodes and edges `_population` and `_properties` are accessed several times, there can be a small performance improvement if they are cached.

For a given circuit, the returned objects should always be the same, and the memory occupation should be minimal compared to the circuit data, so caching them shouldn't be an issue.

Some basic tests:
```python
import bluepysnap
path = "/gpfs/bbp.cscs.ch/project/proj30/tickets/BLPY-289_improve_get/circuit/circuit_config.json"
c = bluepysnap.Circuit(path)
nodes = c.nodes["root__neurons"]

%time nodes._population
%time nodes._properties
%time for i in range(1000): nodes._population
%time for i in range(1000): nodes._properties
```

Results @master:
```
CPU times: user 1.45 ms, sys: 0 ns, total: 1.45 ms
Wall time: 1.48 ms
CPU times: user 7 µs, sys: 13 µs, total: 20 µs
Wall time: 23.1 µs
CPU times: user 156 ms, sys: 105 ms, total: 260 ms
Wall time: 266 ms
CPU times: user 1.64 ms, sys: 0 ns, total: 1.64 ms
Wall time: 1.65 ms
```

Results @more_cached_properties
```
CPU times: user 841 µs, sys: 632 µs, total: 1.47 ms
Wall time: 1.5 ms
CPU times: user 14 µs, sys: 24 µs, total: 38 µs
Wall time: 42.4 µs
CPU times: user 77 µs, sys: 0 ns, total: 77 µs
Wall time: 81.3 µs
CPU times: user 78 µs, sys: 0 ns, total: 78 µs
Wall time: 80.8 µs
```

### DIFF
--- a/bluepysnap/edges/edge_population.py
+++ b/bluepysnap/edges/edge_population.py
@@ -60,12 +60,20 @@ class EdgePopulation:
         self._circuit = circuit
         self.name = population_name
 
-    @property
+    @cached_property
     def _properties(self):
+        """Edge population properties.
+
+        Cached for some performance improvement when it's accessed several times.
+        """
         return self._circuit.to_libsonata.edge_population_properties(self.name)
 
-    @property
+    @cached_property
     def _population(self):
+        """Libsonata edge population.
+
+        Cached for some performance improvement when it's accessed several times.
+        """
         return self._circuit.to_libsonata.edge_population(self.name)
 
     @staticmethod
@@ -590,10 +598,9 @@ class EdgePopulation:
                 )
             ) from e
 
-        properties = self._circuit.to_libsonata.edge_population_properties(self.name)
-        if not properties.spatial_synapse_index_dir:
+        if not self._properties.spatial_synapse_index_dir:
             raise BluepySnapError(f"It appears {self.name} does not have synapse indices")
-        return open_index(properties.spatial_synapse_index_dir)
+        return open_index(self._properties.spatial_synapse_index_dir)
 
     def __getstate__(self):
         """Make EdgePopulation pickle-able, without storing state of caches."""
--- a/bluepysnap/edges/edge_population.py
+++ b/bluepysnap/edges/edge_population.py
@@ -60,19 +60,19 @@ class EdgePopulation:
         self._circuit = circuit
         self.name = population_name
 
-    @cached_property
+    @property
     def _properties(self):
         """Edge population properties.
 
-        Cached for some performance improvement when it's accessed several times.
+        Not cached because quick enough, but do not call in heavy loops.
         """
         return self._circuit.to_libsonata.edge_population_properties(self.name)
 
-    @cached_property
+    @property
     def _population(self):
         """Libsonata edge population.
 
-        Cached for some performance improvement when it's accessed several times.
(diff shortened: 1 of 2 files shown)


joni-herttuainen:
Since this was discussed earlier with @mgeplf in another [PR](https://github.com/BlueBrain/snap/pull/154/files/e8bd639ceaa02c271f02ccd0032889c9fcae4f63#r928722893), I let him to make the call whether we should use `cached_property` or not.

GianlucaFicarelli:
Sure, I can also add more context: in https://github.com/BlueBrain/snap/pull/208 the node population would be retrieved each time `get()` is called, to access `select_all()`, `attribute_names`, and `dynamics_attribute_names`.

In a script like the one provided in https://github.com/BlueBrain/snap/issues/195, `get()` is called multiple times, so the times would sum up when using the code in https://github.com/BlueBrain/snap/pull/208. I needed to profile the code to find where the time was spent, because I didn't expect it was in the `_population` property.

In real use cases, I think that `get()` isn't called so many times, but it seems safer and harmless to use a cached_property even in this case.

bluepysnap/edges/edge_population.py
```diff
@@ -60,12 +60,20 @@ def __init__(self, circuit, population_name):
...
+        """
         return self._circuit.to_libsonata.edge_population_properties(self.name)
 
-    @property
+    @cached_property
     def _population(self):
+        """Libsonata edge population.
+
+        Cached for some performance improvement when it's accessed several times.
+        """
```
mgeplf, line 76: won't this potentially keep a handle to the hdf5 file open? I'm a bit afraid of that, because other hdf5 consumers may exist in the same process, and then errors may be emitted

could you check?

GianlucaFicarelli: You're right, the libsonata population is keeping the h5 file open for reading. I verified with:
```python
import psutil
import libsonata
path = "/gpfs/bbp.cscs.ch/project/proj12/SIT/thalamus_sonata/circuit_sonata.json"
pop = libsonata.CircuitConfig.from_file(path).node_population("thalamus_neurons")
p = psutil.Process()
for i in p.open_files():
    print(i)
```
and I can see, unless pop is deleted:
```
popenfile(path='/gpfs/bbp.cscs.ch/project/proj12/SIT/thalamus_sonata/networks/nodes/thalamus_neurons/nodes.h5', fd=11, position=0, mode='r', flags=32768)
```
However, do you know how it's possible to reproduce errors when multiple hdf5 readers exist in the same process?

mgeplf: > However, do you know how it's possible to reproduce errors when multiple hdf5 readers exist in the same process?

Unfortunately no; in theory, I think multiple readers are safe, but I don't know if that depends on how hdf5 is compiled, or whether there are any other gotchas.

GianlucaFicarelli: ic, I reverted the changes and added a small explanation in the docstring.
I also used self. _properties to simplify a bit the code in spatial_segment_index and spatial_synapse_index

PR #212, opened 2023-05-23
Publish version as bluepysnap.__version__`

GianlucaFicarelli:
Use setuptools_scm to write the version to file.

An alternative could be to use something like:
```python
from importlib.metadata import version

__version__ = version(__package__)
```

but it requires `importlib_metadata` on python 3.7, and it wasn't returning the right version on some local repository (needing more investigation).


joni-herttuainen (approved):
Looks good to me. 👍

PR #213, opened 2023-05-26
Use libsonata to resolve node sets

joni-herttuainen:


bluepysnap/simulation.py
```diff
@@ -126,7 +126,7 @@ def simulator(self):
     def node_sets(self):
```
joni-herttuainen, line 126: Frankly, I am not sure if this should return the `NodeSets` object or the `NodeSets.content`. 

If the use case is resolving `Simulation`'s node sets, then `NodeSets`, if it's just for information, then content, I quess.

bluepysnap/node_sets.py
```diff
@@ -111,12 +40,15 @@ def __init__(self, filepath):
             NodeSets: A NodeSets object.
         """
         self.content = utils.load_json(filepath)
-        self.resolved = _resolve(self.content)
+        self._instance = libsonata.NodeSets.from_file(filepath)
 
-    def __getitem__(self, node_set_name):
+    def resolve_ids(self, node_set_name, population):
```
edasubert, line 45: I am not sure if `resolve_ids` is an appropriate name, I would expect to call `resolve` to resolve internally and then `get` to get the actual value; so I think in this case calling it `get_ids` straight away might be a good choice

joni-herttuainen: yeah, valid point. `get_ids` or `ids` is better

bluepysnap/nodes/node_population.py
```diff
@@ -269,14 +269,16 @@ def _get_node_set(self, node_set_name):
...
 
     def _resolve_nodesets(self, queries):
         def _resolve(queries, queries_key):
             if queries_key == query.NODE_SET_KEY:
                 if query.AND_KEY not in queries:
                     queries[query.AND_KEY] = []
-                queries[query.AND_KEY].append(self._get_node_set(queries[queries_key]))
+                queries[query.AND_KEY].append(
+                    {query.NODE_ID_KEY: self._get_node_set(queries[queries_key])}
+                )
```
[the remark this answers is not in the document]
GianlucaFicarelli: Very good that the times are better, but I'm surprised that querying on Layer23Inhibitory is taking 1 second.

I did a quick check, and I see that most of the time is spent in doing 9 times (one for each mtype) something like:
```python
%time np.in1d(cells['mtype'], 'L23_SBC')
```
that takes around 110ms:
```
CPU times: user 94.2 ms, sys: 17.5 ms, total: 112 ms
Wall time: 111 ms
array([False, False, False, ..., False, False, False])
```
and it's not optimal because:
- mtype is a category, but using np.in1d requires a conversion to object (35ms), and then looking through strings instead of numbers (65ms)
- it should be called only once, passing all the mtypes as a list

For comparison, this calls 
```python
 %time cells['mtype'].isin(['L23_SBC', 'L23_BTC', 'L23_NGC','L23_MC','L23_NBC','L23_LBC','L23_BP','L23_CHC', 'L23_DBC']).to_numpy()
```
would require only 9ms to check all the 9 mtypes at the same time:
```
CPU times: user 9.02 ms, sys: 3 µs, total: 9.02 ms
Wall time: 8.44 ms
array([False, False, False, ..., False, False, False])
```

However, these considerations are unrelated to the MR. I can see if the np.in1d can be safely changed to make the query function more performant, and maybe we can repeat the comparison.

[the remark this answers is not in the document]
joni-herttuainen: Thanks for profiling. 

> I did a quick check, and I see that most of the time is spent in doing 9 times (one for each mtype) something like:
> ```python
> %time np.in1d(cells['mtype'], 'L23_SBC')
> ```

I am not surprised. I guess there has not been a reason to optimize the queries. And, to be fair, 1s for a query is still okayish provided how much data it filters. Obviously, if we can easily reduce it to tenth of that, we should.

> I can see if the np.in1d can be safely changed to make the query function more performant, and maybe we can repeat the comparison.

Am I assuming correctly that you would rather keep the node set resolution in SNAP?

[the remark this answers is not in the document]
GianlucaFicarelli: > Am I assuming correctly that you would rather keep the node set resolution in SNAP?

Nope, but it can be worth to improve the query function because it's used also when `get()` is called with generic queries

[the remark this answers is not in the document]
joni-herttuainen: Ah, of course! Makes sense.

tests/data/node_sets.json
```diff
@@ -5,9 +5,6 @@
   "Layer23": {
     "layer": [2,3]
   },
-  "Empty_nodes": {
-    "node_id": []
-  },
```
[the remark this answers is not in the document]
mgeplf: FYI: I should be closing that soon; we've cleared up the expected semantics.

bluepysnap/edges/edges.py
```diff
@@ -46,7 +46,7 @@ def population_names(self):
         """Defines all sorted edge population names from the Circuit."""
         return sorted(self._circuit.to_libsonata.edge_populations)
 
-    def ids(self, group=None, sample=None, limit=None):
+    def ids(self, group=None, sample=None, limit=None, **_):  # pylint: disable=arguments-differ
```
joni-herttuainen, line 49: How (if at all) can we handle this without disabling `arguments-differ`?

bluepysnap/frame_report.py
```diff
@@ -240,7 +240,7 @@ def data_units(self):
     @property
     def node_set(self):
         """Returns the node set for the report."""
-        return self.simulation.node_sets[self.to_libsonata.cells]
+        return self.simulation.node_sets.content[self.to_libsonata.cells]
```
joni-herttuainen, line 243: Still open to discussion if this should just return the node set name.

Usage would then be:
```python
sim.circuit.get(report.node_set, node_sets=sim.node_sets)
```

bluepysnap/network.py
```diff
@@ -144,13 +144,13 @@ def _get_ids_from_pop(self, fun_to_apply, returned_ids_cls, sample=None, limit=N
         return res
 
     @abc.abstractmethod
-    def ids(self, group=None, sample=None, limit=None):
+    def ids(self, group=None, sample=None, limit=None, node_sets=None):
         """Resolves the ids of the NetworkObject."""
 
     @abc.abstractmethod
-    def get(self, group=None, properties=None):
+    def get(self, group=None, properties=None, node_sets=None):  # pylint: disable=too-many-locals
```
joni-herttuainen, line 151: I can try to get this with less locals in this PR, or we can review it when we make change the get in NSETM-2186.

joni-herttuainen (commented):
@GianlucaFicarelli  Tell me what you think. If this looks somewhat ok, I'll update the docs and add changelog entries. Should we bump this up a major version bearing in mind we will also change `{nodes,edges}.get`?

bluepysnap/frame_report.py
GianlucaFicarelli: From what I see, `cells` is optional at https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#reports

> Specify which node_set to report, default is the simulation “node_set”.

I suppose that returning the name should be enough, but it's better to wait for other opinions.

bluepysnap/edges/edges.py
GianlucaFicarelli: Still for discussion, I was also considering another option that wouldn't require to change the signature of ids() and get().
It would be possible to define a class `NodeSet` that can be returned by `node_sets[node_set_name]`, and that can be passed as `group` to ids() and get().

So, instead of doing
```python
sim.circuit.get(name, node_sets=sim.node_sets)
```
you can override the circuit node_sets for example in this way:
```python
sim.circuit.get(sim.node_sets[name])
```
then get() can internally use the NodeSet to select the ids.

However, there would be a limitation, because in this way it's not possible to specify at the same time a custom node_sets and a query dict containing a node set name, for example when the query is:
```python
{"$node_set": "Node12_L6_Y", "node_id": 1}
```

joni-herttuainen: yeah, we can do something like
```python
sim.circuit.get({
    "$node_set": "Node12_L6_Y", 
    "$node_sets": sim.node_sets,  # probably better to have a more distinctive name like "$node_sets_instance" 
})
```

mgeplf: is that what you were thinking @GianlucaFicarelli? I thought there might be a more clean solution.

GianlucaFicarelli: When filtering by just a single node_set (from a node_sets that's not from the circuit, but from the simulation), I was thinking to:
```python
node_set = sim.node_sets["Node12_L6_Y"]
sim.circuit.get(node_set)
```
If the node_sets come from an external file (not circuit, not simulation), it could be:
```python
node_sets = NodeSets.from_file(...) # or NodeSets(...)
node_set = node_sets["Node12_L6_Y"]
sim.circuit.get(node_set)
```
We can still keep the usual shorter way (when the node_sets is circuit.node_sets):
```python
sim.circuit.get("Node12_L6_Y")
```

If a query is needed when using a different node_set at the same time, something like:
```python
node_set = sim.node_sets["Node12_L6_Y"]
sim.circuit.get({"$node_set": node_set, "morphology": "morph-B")})
```
The downside of this query is that it cannot be dumped as regular json, but maybe this is not desired or required.

About using different keys for node_set name and node_set instance, I'm not convinced that it would make the API easier, but it can be discussed.

bluepysnap/nodes/node_population.py
```diff
@@ -265,18 +266,15 @@ def _check_property(self, prop):
...
-        return self._node_sets[node_set_name]
-
-    def _resolve_nodesets(self, queries):
+    def _resolve_nodesets(self, queries, raise_missing_prop):
         def _resolve(queries, queries_key):
             if queries_key == query.NODE_SET_KEY:
                 if query.AND_KEY not in queries:
                     queries[query.AND_KEY] = []
-                queries[query.AND_KEY].append(self._get_node_set(queries[queries_key]))
+                node_set = self._node_sets[queries[queries_key]]
```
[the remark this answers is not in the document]
GianlucaFicarelli: Yes this one. I don't have a strong opinion either, let's see if there are other opinions or known use cases.

bluepysnap/node_sets.py
```diff
@@ -111,12 +67,14 @@ def __init__(self, filepath):
             NodeSets: A NodeSets object.
         """
         self.content = utils.load_json(filepath)
-        self.resolved = _resolve(self.content)
+        self._instance = libsonata.NodeSets.from_file(filepath)
```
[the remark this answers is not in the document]
GianlucaFicarelli: If we consider public the NodeSets class, I'd like if both snap and libsonata allow to init the object from str, path, dict.

libsonata already allows:
- `libsonata.NodeSets(<content>)`
- `libsonata.NodeSets.from_file(<path>)`

so it's missing only the init from a dict. Was it discussed before @mgeplf ?

If we want, we can still wrap around libsonata if we define in snap:
- `NodeSets.from_file()`
- `NodeSets.from_dict()`
- `NodeSets.from_string()`

or we can avoid one of them, if we define the init accordingly.

These methods probably don't have a lot of applications, but the main use case could be to build custom node_sets in memory (as I would like to do if it were possible to convert a query group to node_set, as I was trying to do in another PR).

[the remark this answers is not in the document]
mgeplf: > Was it discussed before @mgeplf ?

No; I'm a bit sad I did the `libsonata.NodeSets(<content>)`, so I think having it clear w/ `from_file()`, `from_string()` makes good sense.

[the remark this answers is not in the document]
joni-herttuainen: Added all three suggested methods

bluepysnap/node_sets.py
```diff
@@ -20,82 +20,38 @@
...
-        _resolve_set(content, resolved, set_name)
-    return resolved
+    def get_ids(self, population, raise_missing_property=True):
+        """Get the resolved node set as ids."""
+        try:
+            return self._node_sets.materialize(self._name, population).flatten()
+        except libsonata.SonataError as e:
+            if "No such attribute" in e.args[0]:
+                if not raise_missing_property:
+                    return []
```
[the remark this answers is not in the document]
joni-herttuainen: thanks. Tunnel-visioned big time there.

bluepysnap/simulation.py
```diff
@@ -126,7 +126,7 @@ def simulator(self):
...
         """Returns the NodeSets object bound to the simulation."""
         node_sets_file = self.to_libsonata.node_sets_file
-        return NodeSets(node_sets_file) if node_sets_file else {}
+        return NodeSets.from_file(node_sets_file) if node_sets_file else {}
```
GianlucaFicarelli, line 129: This wasn't changed in this PR, but it would be more consistent if we always return a NodeSet instance, instead of returning an empty dict when `node_sets_file` is empty.
What about of something like this?
```python
        return NodeSets.from_file(node_sets_file) if node_sets_file else NodeSets.from_string("{}")
```
or 
```python
        return NodeSets.from_file(node_sets_file) if node_sets_file else NodeSets.from_dict({})
```

doc/source/notebooks/03_node_properties.ipynb
```diff
@@ -186,14 +186,12 @@
...
       "text/plain": [
-       "<Figure size 432x288 with 1 Axes>"
+       "<Figure size 640x480 with 1 Axes>"
```
GianlucaFicarelli, line 191: Just wondering, why images are now 640x480 instead of 432x288?

doc/source/notebooks/03_node_properties.ipynb
```diff
@@ -310,16 +317,22 @@
    "execution_count": 5,
    "metadata": {},
    "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/gpfs/bbp.cscs.ch/data/scratch/proj30/home/herttuai/NSETM_2185/venv/lib/python3.10/site-packages/pandas/plotting/_matplotlib/core.py:1256: UserWarning: No data for colormapping provided via 'c'. Parameters 'cmap' will be ignored\n",
```
GianlucaFicarelli, line 324: This warning is new, is it possible to fix it? (here or in a different PR)

doc/source/notebooks/03_node_properties.ipynb
```diff
@@ -481,35 +493,37 @@
...
...ab693vfHAPiv5rYUfwxglWVZewF8CdxtcR+y3WYwxv7d+zsG7tc8C9nuF78A8AvG2Pe8/78CrqCz3GbCnwJ4ijH2K+//dmhzFYxCTgGWZVkAHgTwLGPsbuGrbwJY731eD+5bpu3rvNXecwDsF6ZWzWjvYsuyjvU+l8D93c+CK+a3KtpLv+OtAP7RsziaBsbYBxhjL2eMLQWflv4jY+xqZLjNlmWVLcvqpM/g/s0fIaP9AgAYY/8J4OeWZZ3sbboQwGiW2yxgDXx3BdAeba5Gq53YC0EAnAs+HXoGwNOevBHcZ/kYgJ8A2AXgJd7+FoD7wf22PwRwRpPb+4cAdnvt/RGA27ztrwQwAmAP+LTP9bYXvf/3eN+/ssX3+3wA38p6m722/cCTHwP4oLc9k/1CaPdpAJ7w+sfXARzXBm0ug8+AjhG2ZbrNMjGZegYGBgYZgXFZGBgYGGQERiEbGBgYZARGIRsYGBhkBEYhGxgYGGQERiEbGBgYZARGIRsYeLAsq2RZ1ncty8qncK7FlmX9QxrtMjh6YBSygYGPDQC+xhibrfdEjLFxAL+0LOuP62+WwdECo5ANFjwsyzrT470tetlzP7Ys6w8ku14NL5vLsqzzLY9z2fv/U5Zlvd37vNeyrI8Sx7FlWcstyxqyLOunlmVdJ5zv6945DQy0UIjexcCgvcEY+75lWd8EcDuAEoCHGWM/EvfxiGleyRjbq3nanzHGTrMs6x4A28G5NorgmY8D3j5PeNc0MNCCUcgGRws+AuD7ACYAbJZ8vwjAizHO903v7w8BVBhjBwAcsCxr0rKsYxknbRoD8NuJW2xw1MG4LAyOFrwUQAVAJ7glG8QRyXZL+GwHvpv0/s4Jn+l/MnSK3nkNDLRgFLLB0YIHAPwFgC9CUs6JMfYCgLxlWaJS/gMv8sIFpyeNG33xu+AuDAMDLRiFbLDgYVnWOgDTjLEdAD4G4EzLst4g2fU74Mx9hBfAS0H9Kzhb2Ac8Gk1dXADgfyVrtcHRCMP2ZmDgwas+fCNj7M8syzofwM2MsTfXcb7/DeBSz/o2MIiEsZANDDwwxp4C8HhaiSEA7jbK2CAOjIVsYGBgkBEYC9nAwMAgIzAK2cDAwCAjMArZwMDAICMwCtnAwMAgIzAK2cDAwCAjMArZwMDAICP4/wF7b5YiC9McPgAAAABJRU5ErkJggg==\n",
```
GianlucaFicarelli, line 502: If I'm not wrong, this image isn't present in the modified notebook

bluepysnap/simulation.py
GianlucaFicarelli: Alternatively, we could return None, but it seems a bigger breaking change.

doc/source/notebooks/03_node_properties.ipynb
joni-herttuainen: Not sure. I mean the default seems to have been 640x480 for a while:
https://matplotlib.org/stable/users/prev_whats_new/dflt_style_changes.html#figure-size-font-size-and-screen-dpi

So the question is more like why was it 432x288?

doc/source/notebooks/03_node_properties.ipynb
joni-herttuainen: I added code to temporarily ignore it since I didn't find an immediate solution.

doc/source/notebooks/03_node_properties.ipynb
joni-herttuainen: Yeah, sometimes when I run it on a local (well, inside bb5) jupyter instance, it skips a cell. this is what happened here. Should be fixed now

mgeplf:
Things look reasonable to me, but I haven't been following it as closely as @GianlucaFicarelli - IMO he should have the final say.

bluepysnap/simulation.py
GianlucaFicarelli: Good, this is fixed in `Circuit.node_sets`, but it should be changed in `Simulation.node_sets` as well (and the changelog updated to mention Simulation)

joni-herttuainen: 🤦‍♂️ , I remembered that when there is no simulation node set file defined, libsonata returns circuits node set path so this needed no fixing. OFC, both can be empty. 

Thanks > should be fixed now.

GianlucaFicarelli (commented):
LGTM 👍 

Should we release 1.0.7 before merging and releasing 1.1.0?
Alternatively, the changelog should be updated (moving the entries from 1.0.7 to 1.1.0)

joni-herttuainen:
I'm happy with merging and releasing any none-breaking changes first

GianlucaFicarelli:
> I'm happy with merging and releasing any none-breaking changes first

I'd like to merge https://github.com/BlueBrain/snap/pull/220 once it's approved (since it's not a breaking change), release 1.0.7, then merge this PR and release 1.1.0. What do you think?

joni-herttuainen:
Sounds good. I think #220 is good to go.

bluepysnap/nodes/node_population.py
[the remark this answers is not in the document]
GianlucaFicarelli: Do you mind repeating the same benchmark with version 1.0.7, after merging or rebasing on master?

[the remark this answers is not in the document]
joni-herttuainen: with `snap`:
```
# ROUNDS: 100

# Layer23Inhibitory #
- total time: 5.96 s
- per round (avg): 0.060s

# S1HL #
- total time: 2.55 s
- per round (avg): 0.025s

# mc2;VPL #
- total time: 0.66 s
- per round (avg): 0.007s

# mc2_Column #
- total time: 6.51 s
- per round (avg): 0.065s

--------------------------------
# ROUNDS: 1000

# Layer23Inhibitory #
- total time: 27.48 s
- per round (avg): 0.027s

# S1HL #
- total time: 19.42 s
- per round (avg): 0.019s

# mc2;VPL #
- total time: 6.56 s
- per round (avg): 0.007s

# mc2_Column #
- total time: 66.19 s
- per round (avg): 0.066s
```

with `libsonata`:
```
# ROUNDS: 100

# Layer23Inhibitory #
- total time: 3.39 s
- per round (avg): 0.034s

# S1HL #
- total time: 9.93 s
- per round (avg): 0.099s

# mc2;VPL #
- total time: 3.95 s
- per round (avg): 0.040s

# mc2_Column #
- total time: 7.02 s
- per round (avg): 0.070s

--------------------------------
# ROUNDS: 1000

# Layer23Inhibitory #
- total time: 26.13 s
- per round (avg): 0.026s

# S1HL #
- total time: 103.41 s
- per round (avg): 0.103s

# mc2;VPL #
- total time: 40.42 s
- per round (avg): 0.040s

# mc2_Column #
- total time: 71.51 s
- per round (avg): 0.072s
```

GianlucaFicarelli (approved):
LGTM

PR #216, opened 2023-06-01
Produce empty plot when there are no spikes

### LINKED ISSUE #214: firing_rate_histogram is failing when there are no spikes
The desired behaviour is to produce empty plot without any errors. Thanks for taking a look!

joni-herttuainen:
Fixes #214


GianlucaFicarelli (approved):
I wonder if we can add tests for _plotting, but the PR LGTM

joni-herttuainen:
Yeah, it might be a bit tricky but possible

PR #218, opened 2023-06-08
Add t_step parameter to frame reports

GianlucaFicarelli:
See NSETM-2194

### DIFF
--- a/bluepysnap/frame_report.py
+++ b/bluepysnap/frame_report.py
@@ -85,6 +85,8 @@ class PopulationFrameReport:
             t_stop (float): Include only frames occurring at or before this time.
             t_step (float): Optional time step, useful to reduce the number of samples.
                 It should be a multiple of the report time step dt, and it's equal to dt by default.
+                If the given t_step isn't an exact multiple, it's rounded to the closer multiple.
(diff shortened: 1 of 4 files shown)


joni-herttuainen (approved):
:metal:

bluepysnap/frame_report.py
```diff
@@ -76,20 +76,28 @@ def _wrap_columns(columns):
...
 
-    def get(self, group=None, t_start=None, t_stop=None):
+    def get(self, group=None, t_start=None, t_stop=None, t_step=None):
         """Fetch data from the report.
 
         Args:
             group (None/int/list/np.array/dict): Get frames filtered by :ref:`Group Concept`.
             t_start (float): Include only frames occurring at or after this time.
             t_stop (float): Include only frames occurring at or before this time.
+            t_step (float): Optional time step, useful to reduce the number of samples.
```
mgeplf, line 86: I think it's important that this explains exactly what sort of stepping is done; ie, is it averaging skipped samples, etc.

GianlucaFicarelli: Do you think it's correct and clear enough in this way? Suggestions are welcome :)

mgeplf: yup, looks good

PR #220, opened 2023-06-09
Improve performance when querying a population with get() and ids()

GianlucaFicarelli:
Main improvements:
- use pandas series when the type is category, instead of converting to numpy arrays of objects
- avoid creating full numpy arrays of True or False
- quickly return when False or True is found in masks, if possible

See NSETM-2181 for more details.

### DIFF
--- a/bluepysnap/query.py
+++ b/bluepysnap/query.py
@@ -1,6 +1,8 @@
 """Module to process search queries of nodes/edges."""
+import operator
 from collections.abc import Mapping
 from copy import deepcopy
+from functools import reduce
 
 import numpy as np
 
@@ -19,12 +21,51 @@ VALUE_KEYS = {REGEX_KEY}
 ALL_KEYS = {NODE_ID_KEY, EDGE_ID_KEY, POPULATION_KEY, OR_KEY, AND_KEY, NODE_SET_KEY} | VALUE_KEYS
 
 
-# TODO: move to `libsonata` library
+def _logical_and(masks):
+    """Calculate and return the "logical and" of the given masks.
+
+    Args:
+        masks: list of array-like, or booleans.
+
+    Returns:
+        the resulting mask as a numpy array, or a bool. Return True if the masks list is empty.
+    """
+    # filter out True masks (as bool)
+    masks = [m for m in masks if m is not True]
+    if any(m is False for m in masks):
+        return False
+    if len(masks) == 0:
+        return True
+    masks = (np.asarray(m, dtype=bool) for m in masks)
+    # faster than np.logical_and.reduce or np.all
+    return reduce(operator.and_, masks)
+
+
+def _logical_or(masks):
+    """Calculate and return the "logical or" of the given masks.
+
+    Args:
+        masks: list of array-like, or booleans.
+
+    Returns:
+        the resulting mask as a numpy array, or a bool. Return False if the masks list is empty.
+    """
+    # filter out False masks (as bool)
+    masks = [m for m in masks if m is not False]
+    if any(m is True for m in masks):
+        return True
+    if len(masks) == 0:
+        return False
+    masks = (np.asarray(m, dtype=bool) for m in masks)
+    # faster than np.logical_or.reduce or np.any
+    return reduce(operator.or_, masks)
+
+
 def _complex_query(prop, query):
-    result = np.full(len(prop), True)
+    result = True
     for key, value in query.items():
         if key == REGEX_KEY:
-            result = np.logical_and(result, prop.str.match(value + "\\Z"))
+            result = _logical_and([result, prop.str.match(value + "\\Z")])
         else:
             raise BluepySnapError(f"Unknown query modifier: '{key}'")
     return result
@@ -41,7 +82,7 @@ def _positional_mask(data, ids):
         _positional_mask(data, [0,2]) --> [True, False, True, False, False]
     """
     if ids is None:
-        return np.full(len(data), fill_value=True)
+        return True
     if isinstance(ids, int):
         ids = [ids]
     mask = np.full(len(data), fill_value=False)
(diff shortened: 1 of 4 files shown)


GianlucaFicarelli:
Some benchmarks about reducing masks with OR:
```python
import operator
import numpy as np
from functools import reduce

def _wrap_or(x, y):
    return operator.or_(x, y)

np.random.seed(0)
masks = [np.random.choice([True, False], 80_000_000) for i in range(6)]

for i in range(1, 6):
    print(i)
    _ = %timeit reduce(operator.or_, masks[:i])
    _ = %timeit reduce(np.logical_or, masks[:i])
    _ = %timeit reduce(_wrap_or, masks[:i])
    _ = %timeit np.any(masks[:i], axis=0)
    _ = %timeit np.logical_or.reduce(masks[:i])
```
Result:
```
1
335 ns ± 4.62 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
334 ns ± 0.0464 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
296 ns ± 0.0496 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
67.2 ms ± 57.9 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
67.1 ms ± 44.6 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
2
30.9 ms ± 52.4 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
30.9 ms ± 49.2 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
30.9 ms ± 32.3 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
113 ms ± 442 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
112 ms ± 214 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
3
43.4 ms ± 53.8 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
63.1 ms ± 23 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
63.2 ms ± 32 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
157 ms ± 942 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
159 ms ± 326 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
4
56.6 ms ± 18.7 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
96.8 ms ± 41.2 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
96.8 ms ± 55.1 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
204 ms ± 2.24 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
204 ms ± 896 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
5
69.3 ms ± 57.7 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
129 ms ± 38.5 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
127 ms ± 30.8 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
247 ms ± 411 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
247 ms ± 807 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Calling `reduce(operator.or_, masks[:i])` is almost always faster.
Looking at the flame graph, I think that in this case numpy is able to use the elision code https://github.com/numpy/numpy/blob/c6e1fd039da2e6127ba0798f327e9e1b122d7534/numpy/core/src/multiarray/temp_elide.c
from https://github.com/numpy/numpy/blob/c6e1fd039da2e6127ba0798f327e9e1b122d7534/numpy/core/src/multiarray/number.c#L592
while in the other cases it's not.

bluepysnap/query.py
```diff
@@ -19,12 +21,57 @@
...
+    filtered_masks = []
+    for mask in masks:
+        if mask is True:
+            return True
+        if mask is False:
+            continue
+        filtered_masks.append(mask)
+    if not filtered_masks:
+        return False
+    # reduce(operator.or_, masks) is faster than np.logical_or.reduce() or np.any()
```
mgeplf, line 66: huh; fun

bluepysnap/query.py
```diff
@@ -19,12 +21,57 @@
...
+    return reduce(operator.or_, (np.asarray(m, dtype=bool) for m in filtered_masks))
+
+
 def _complex_query(prop, query):
-    result = np.full(len(prop), True)
+    result = True
     for key, value in query.items():
         if key == REGEX_KEY:
-            result = np.logical_and(result, prop.str.match(value + "\\Z"))
+            result = _logical_and([result, prop.str.match(value + "\\Z")])
```
mgeplf, line 74:
```suggestion
            result = _logical_and([result, prop.str.match(value + r"\Z")])
```

mgeplf (commented):
seems very nice/reasonable to me

bluepysnap/query.py
```diff
@@ -19,12 +21,57 @@
...
+    for mask in masks:
+        if mask is True:
+            return True
+        if mask is False:
+            continue
+        filtered_masks.append(mask)
+    if not filtered_masks:
+        return False
+    # reduce(operator.or_, masks) is faster than np.logical_or.reduce() or np.any()
+    return reduce(operator.or_, (np.asarray(m, dtype=bool) for m in filtered_masks))
```
joni-herttuainen, line 67: Wow, this is lightning fast, tried a few things but just couldn't beat this. Good job!

joni-herttuainen (approved):
If it was me, I'd add some sanity checks just for the logical functions to check that:
`_logical_and` == `np.logical_and`, and 
`_logical_or`  == `np.logical_or`
but given how many tests they go through already, this feels almost redundant.

GianlucaFicarelli:
> If it was me, I'd add some sanity checks just for the logical functions to check that:

Sure, I'll add some explicit tests for _logical_or and _logical_and.

joni-herttuainen (approved):

PR #222, opened 2023-06-14
Fix CircuitIds.sample() to always return different samples

GianlucaFicarelli:


joni-herttuainen (approved):
well spotted

GianlucaFicarelli:
I've found it when looking at MR 13 of minis-validation, cc @edasubert 
Note that if ids.sample() was used in tests in other projects, it may return different ids even if the seed was initialized, so the tests may fail.

PR #227, opened 2023-06-27
fix warning in a notebook

joni-herttuainen:
Fixed the warning in a notebook when doing scatter plot that we temporarily suppressed in #213 .


joni-herttuainen:
TBH, I still don't fully understand how and why this warning is raised but giving the color in an array seems to fix it. I think it was introduced in https://github.com/pandas-dev/pandas/pull/48734#pullrequestreview-1119104409  but other than that... 🤷‍♂️

doc/source/notebooks/03_node_properties.ipynb
```diff
@@ -333,11 +333,8 @@
     "\n",
     "df_grouped = df_all.groupby('synapse_class')\n",
     "\n",
-    "import warnings\n",
-    "with warnings.catch_warnings():\n",
-    "    warnings.simplefilter(\"ignore\") # For: UserWarning: No data for colormapping provided via 'c'.\n",
-    "    for color, (name, group) in zip(('r', 'b'), df_grouped):\n",
-    "        group.plot.scatter(x='x', y='z', c=color, ax=ax, label=name)\n",
+    "for color, (name, group) in zip(('r', 'b'), df_grouped):\n",
+    "    group.plot.scatter(x='x', y='z', c=[color], ax=ax, label=name)\n",
```
GianlucaFicarelli, line 337: The warning seems like a bug with Pandas to me.

I noticed that it goes away not only if you pass the color as a list, but also when you pass the long string of the colors ('red', 'blue' instead of 'r', 'b'), without wrapping with a list.

According to the docs https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.scatter.html, "A single color string referred to by name, RGB or RGBA code, for instance ‘red’ or ‘#a98d19’" should be valid.

I'd suggest also to reduce the size of the dataframe using `sample`, to make the graph more meaningful (right now, the blue INH points completely overlap the red EXC points). Using 5000 rows, and size=3, just as an example, I find it more clear:

```python
df_grouped = df_all.sample(5000, random_state=0).groupby('synapse_class')

for color, (name, group) in zip(('red', 'blue'), df_grouped):
    group.plot.scatter(x='x', y='z', s=3, c=color, ax=ax, label=name)

```

GianlucaFicarelli: I found out that Pandas is passing a colormap instead of None when `c='b'`, and this is causing the warning in matplotlib, because the colormap is ignored.
This is indirectly caused by `b` considered as dtype int8 (I wonder if this is expected):
```python
>>> from pandas.core.dtypes.common import is_integer_dtype, pandas_dtype
>>> is_integer_dtype("b")
True
>>> pandas_dtype("b")
dtype('int8')
```
If `blue` or a list is passed, the colormap isn't passed and the warning isn't generated.

joni-herttuainen: Interesting. I, too, wonder, if this is a bug or an intended feature 🤔

joni-herttuainen: Went with 30k and s=1. I created a variable for them so it is easier to adjust it.

GianlucaFicarelli (approved):

PR #229, opened 2023-07-17
Add query key `population_type`

### LINKED ISSUE #228: Filter by node or edge population `type`
When getting node and edge populations, it should be able possible to filter by `type` (as defined in the config file);
ie: `get all virtual edge populations`, or `only in non-virtual node populations`.
The idea would be that this could happen across the API.

joni-herttuainen:
Added a query key `population_type` to query edge / node ids based on population type.

Examples:
```python
# Nodes
circuit.nodes.ids({"population_type": "biophysical"})
circuit.nodes.get({"population_type": ["astrocyte", "biophysical"]}, properties=["x", "y", "z"])

# Edges
circuit.nodes.ids({"population_type": "electrical"})
# NOTE: circuit.edges.get can't be queried the same way as circuit.nodes.get
```

Closes #228

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Version v3.0.1
+--------------
+
+New Features
+~~~~~~~~~~~~
+- Added the possibility to query Edge IDs and Node IDs based on edge/node population type using query key ``population_type``
+
+
 Version v3.0.0
 --------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,8 @@ New Features
 ~~~~~~~~~~~~
 - Added the possibility to query Edge IDs and Node IDs based on edge/node population type using query key ``population_type``
 
+  - the types conform to `node types <https://sonata-extension.readthedocs.io/en/latest/sonata_config.html#populations>`_ and `edge types <https://sonata-extension.readthedocs.io/en/latest/sonata_config.html#id4>`_ defined in the sonata specification
+
 
 Version v3.0.0
 --------------
--- a/bluepysnap/query.py
+++ b/bluepysnap/query.py
@@ -101,6 +101,8 @@ def _positional_mask(data, ids):
         return True
     if isinstance(ids, int):
         ids = [ids]
+    elif len(ids) == 0:
+        return False
     mask = np.full(len(data), fill_value=False)
     indices = data.index.get_indexer(ids)
     mask[indices[indices > -1]] = True
--- a/bluepysnap/query.py
+++ b/bluepysnap/query.py
@@ -111,9 +111,9 @@ def _circuit_mask(data, population_name, population_type, queries):
     """Handle the population, node ID queries."""
     populations = queries.pop(POPULATION_KEY, None)
     types = queries.pop(POPULATION_TYPE_KEY, None)
-    if populations is not None and population_name not in set(utils.ensure_list(populations)):
+    if populations is not None and population_name not in utils.ensure_list(populations):
         ids = []
-    elif types is not None and population_type not in set(utils.ensure_list(types)):
+    elif types is not None and population_type not in utils.ensure_list(types):
         ids = []
     else:
         ids = queries.pop(NODE_ID_KEY, queries.pop(EDGE_ID_KEY, None))
--- a/bluepysnap/query.py
+++ b/bluepysnap/query.py
@@ -14,12 +14,21 @@ from bluepysnap.exceptions import BluepySnapError
 NODE_ID_KEY = "node_id"
 EDGE_ID_KEY = "edge_id"
 POPULATION_KEY = "population"
+POPULATION_TYPE_KEY = "population_type"
 OR_KEY = "$or"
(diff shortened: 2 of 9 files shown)


bluepysnap/query.py
```diff
@@ -97,23 +106,26 @@ def _positional_mask(data, ids):
...
 
 
-def _circuit_mask(data, population_name, queries):
+def _circuit_mask(data, population_name, population_type, queries):
     """Handle the population, node ID queries."""
     populations = queries.pop(POPULATION_KEY, None)
+    types = queries.pop(POPULATION_TYPE_KEY, None)
     if populations is not None and population_name not in set(utils.ensure_list(populations)):
         ids = []
+    elif types is not None and population_type not in set(utils.ensure_list(types)):
```
GianlucaFicarelli, line 115: Building a set from the list here isn't useful, because it requires to iterate over the list.
Using `population_type not in utils.ensure_list` is enough (the same applies to the line above for populations, even it wasn't changed in this PR).

joni-herttuainen:
> `# NOTE: circuit.edges.get can't be queried the same way as circuit.nodes.get`

Actually, not quite true but we never claim that we'd support it:
```python
circuit.edges.get({"population_type": "chemical"}) 
# returns: {"population_type": "chemical"} (dict)

circuit.edges.get({"population_type": "chemical", properties=["afferent_section_id"]}
# returns: pd.DataFrame with what you'd expect
```
Same happens with other queries used with `Edges.get`. Do we want to do something about this or leave it as is?

CHANGELOG.rst
```diff
@@ -11,7 +11,8 @@ New Features
 Improvements
 ~~~~~~~~~~~~
 - Node set resolution is done by libsonata
-- Added kwarg: `raise_missing_property` to `NodePopulation.get`
+- Added kwarg: ``raise_missing_property`` to ``NodePopulation.get``
+- Possibility to query Edge IDs and Node IDs based on edge/node population type using query key ``population_type``
```
GianlucaFicarelli, line 15: Minor
```suggestion
- Added possibility to query Edge IDs and Node IDs based on edge/node population type using query key ``population_type``
```

bluepysnap/query.py
joni-herttuainen: True, copy-pasta.
-> fixed

bluepysnap/query.py
```diff
@@ -172,7 +184,7 @@ def _collect(_, query_key):
     return props
 
 
-def resolve_ids(data, population_name, queries):
+def resolve_ids(data, population_name, population_type, queries):
     """Returns an index mask of `data` for given `queries`.
 
     Args:
```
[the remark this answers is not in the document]
joni-herttuainen: Done

bluepysnap/query.py
```diff
@@ -97,23 +106,26 @@ def _positional_mask(data, ids):
...
     populations = queries.pop(POPULATION_KEY, None)
-    if populations is not None and population_name not in set(utils.ensure_list(populations)):
+    types = queries.pop(POPULATION_TYPE_KEY, None)
+    if populations is not None and population_name not in utils.ensure_list(populations):
+        ids = []
+    elif types is not None and population_type not in utils.ensure_list(types):
         ids = []
     else:
         ids = queries.pop(NODE_ID_KEY, queries.pop(EDGE_ID_KEY, None))
     return queries, _positional_mask(data, ids)
```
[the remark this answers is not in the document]
joni-herttuainen: Done

CHANGELOG.rst
```diff
@@ -12,6 +12,9 @@ New Features
 
   - deprecated ``validate``
 
+Improvements
+~~~~~~~~~~~~
+- Added possibility to query Edge IDs and Node IDs based on edge/node population type using query key ``population_type``
```
GianlucaFicarelli, line 17: It would be worth to mention that the ``population_type`` corresponds to ``type`` in the [sonata specs](https://sonata-extension.readthedocs.io/en/latest/sonata_config.html).

I also feel that the group concept could be documented more, because there is a "Query system" section in [From bluepy to bluepysnap](https://bbpgitlab.epfl.ch/nse/bluepy/-/blob/main/examples/ipython_notebooks/From%20bluepy%20to%20bluepysnap.ipynb) but it's not public.

For example we could add other examples in [Group Concept](https://github.com/BlueBrain/snap/blob/c9162f7373c79975f5e5232bd2896e9e2b50013e/bluepysnap/nodes/node_population.py#L20) using ``population_type``,``population`` and ``node_id``.

joni-herttuainen: I agree, I think it would also be beneficial to demo the different query options in the notebooks (I'll create a separate PR for that). 

Meanwhile, I'll update the docstring(s) and the change log.

mgeplf:
I think this is good to go, yeah? Seems reasonable to me, I think.

CHANGELOG.rst
joni-herttuainen: I added examples of `population_type` in the notebooks, that were added in #261 .

GianlucaFicarelli (approved):

PR #217, opened 2023-06-08
Fixing dtype issues in network.get

joni-herttuainen:
Fixes #198 .
Fixes #203 .

Furthermore the following can be closed:  
* #199
* #180 
* #186

PR #232, opened 2023-08-31
fix dtypes in CircuitIds.index_schema and several deprecation warnings

joni-herttuainen:
In `pandas 2.1.0` release, something changed to cause 
```python
pd.MultiIndex.from_tuples([], names=self.index.names)
```
to return a MultiIndex with all names having `dtype=object` instead of preserving them, which broke some tests with `pdt.assert_index_equal`.

This is a fix for it.


bluepysnap/circuit_ids.py
```diff
@@ -58,7 +58,11 @@ def __init__(self, index, sort_index=True):
     @property
     def index_schema(self):
         """Return an empty index with the same names of the wrapped index."""
-        return pd.MultiIndex.from_tuples([], names=self.index.names)
+        # Since pandas v2.1.0, need to set dtypes explicitly
+        levels = [np.array([], dtype=dtype) for dtype in self.index.dtypes]
+        codes = len(levels) * [[]]
+
+        return pd.MultiIndex(levels=levels, codes=codes, names=self.index.names)
```
GianlucaFicarelli, line 65: I see that the result of `pd.MultiIndex.from_tuples([], names=["a", "b"])` is the same in pandas 2.0.3 and 2.1.0 (i.e. `idx.dtypes` are always objects).
I suspect that something changed in `assert_index_equal()` or `assert_frame_equal()`.

Regardless, instead of manually building a new index, you could just return:
```python
return self.index[:0]
```

joni-herttuainen: > I see that the result of pd.MultiIndex.from_tuples([], names=["a", "b"]) is the same in pandas 2.0.3 and 2.1.0 (i.e. idx.dtypes are always objects).

Damnit. I'll take another look to get to the bottom of this.

> Regardless, instead of manually building a new index, you could just return

Holy cow, I tried just about everything but not that. 🤦  Thanks!

joni-herttuainen: AAMOF, the position of the multi-index having dtypes of object does not matter: as long as there's any of the indices having dtype=object, the concatenated index will have that dtype.

GianlucaFicarelli: Thanks for the investigation. Another thing that I noticed in your example:

```python
>>> dfs[0]['c'].dtype
dtype('float64')
>>> dfs[1]['c'].dtype
dtype('int64')
>>> df['c'].dtype
dtype('float64')
```

To avoid casting to float64, in cases like this one, we may need to ensure that the dtype of `dfs[0]['c']` is set to `int64` (or skip empty dataframes when concatenating, but it seems less clean)

joni-herttuainen: I believe that the setting of dtypes is covered in https://github.com/BlueBrain/snap/blob/master/bluepysnap/network.py#L169C10-L169C10

Either way, the dataframes will no longer be concatenated after #217 , nor will the `index_schema` be used.

joni-herttuainen:
I fixed pretty much all the deprecation warnings popping up in tests. Few notions:
* removed `{Edges,EdgePopulaiton}.properties` as it was already deprecated and supposed to be removed in v1.0.0
  * I believe the circuit-faker is the only thing using it 
* removed deprecation on calling `{Edges, EdgePopulation}.get` with `properties=None` (returns `ids`) as I saw no particular reason as to why it should not be allowed
* `pkg_resources` -> `importlib_resources` as `importlib.resources.files` is not available in `py38`
  * However in `doc/source/conf.py`: `pkg_resources` -> `importlib` as docs are built with `py310`/`py311`

bluepysnap/spike_report.py
```diff
@@ -167,9 +167,16 @@ def report(self):
...
             data["population"] = np.full(len(data), population)
-            res = pd.concat([res, data])
-        if not res.empty:
-            res["population"] = res["population"].astype("category")
+
+            if not data.empty:
+                dfs.append(data)
+
+        if len(dfs) == 0:
+            return pd.DataFrame()
```
GianlucaFicarelli, line 175: For consistency, it would be better to return an empty dataframe with the same schema (index and columns), instead of a generic empty dataframe, and add a test for it.
We fixed a similar issue for frame reports in https://github.com/BlueBrain/snap/pull/224

bluepysnap/schemas/schemas.py
```diff
@@ -2,9 +2,10 @@
 from pathlib import Path
 
 import h5py
+import importlib_resources
 import jsonschema
 import numpy as np
-import pkg_resources
+import referencing
```
GianlucaFicarelli, line 8: The library `referencing` should be added explicitly as a requirement in setup.py.
I suppose (but I didn't verify) that we should also pin `jsonschema>=4.18.0`, since it's the version that starts using referencing.

mgeplf (approved):

PR #235, opened 2023-10-09
Extend circuit node_sets with simulation node sets

joni-herttuainen:
Also
* Add `update` method to node sets
* restrict `libsonata>=0.1.24`

### DIFF
--- a/bluepysnap/node_sets.py
+++ b/bluepysnap/node_sets.py
@@ -89,6 +89,23 @@ class NodeSets:
         """Create NodeSets instance from a dict."""
         return cls.from_string(json.dumps(content))
 
+    @property
+    def to_libsonata(self):
+        """Libsonata instance of the NodeSets."""
+        return self._instance
+
+    def update(self, node_sets):
+        """Update the contents of the node set."""
+        if not isinstance(node_sets, NodeSets):
+            raise BluepySnapError(
+                f"Unexpected type: '{type(node_sets).__name__}' "
+                f"(expected: '{self.__class__.__name__}')"
+            )
+
+        overwritten = self._instance.update(node_sets.to_libsonata)
+        self.content.update(node_sets.content)
+        return overwritten
+
     def __contains__(self, name):
         """Check if node set exists."""
         if isinstance(name, str):
--- a/bluepysnap/node_sets.py
+++ b/bluepysnap/node_sets.py
@@ -21,6 +21,7 @@ For more information see:
 https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#node-sets-file
 """
 
+import copy
 import json
 
 import libsonata
@@ -67,7 +68,7 @@ class NodeSets:
         Returns:
             NodeSets: A NodeSets object.
         """
-        self.content = content
+        self._content = content
         self._instance = instance
 
     @classmethod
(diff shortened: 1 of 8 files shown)


bluepysnap/node_sets.py
```diff
@@ -89,6 +90,28 @@ def from_dict(cls, content):
...
+        """Access (a copy of) the node sets contents."""
+        return copy.deepcopy(self._content)
+
+    @property
+    def to_libsonata(self):
+        """Libsonata instance of the NodeSets."""
+        return self._instance
+
+    def update(self, node_sets):
+        """Update the contents of the node set."""
```
GianlucaFicarelli, line 104: It should be mentioned in the docstring that `update` returns a set with the names of overwritten node_sets.

bluepysnap/simulation.py
```diff
@@ -125,8 +136,17 @@ def simulator(self):
...
-        return NodeSets.from_file(path) if path else NodeSets.from_dict({})
+        try:
+            path = self.circuit.to_libsonata.node_sets_path
+        except BluepySnapError:
+            path = ""
+
+        node_sets = NodeSets.from_file(path) if path else NodeSets.from_dict({})
+
+        if path := self.to_libsonata.node_sets_file:
+            _warn_on_overwritten_node_sets(node_sets.update(NodeSets.from_file(path)))
```
GianlucaFicarelli, line 147: For clarity I would split this in 2 lines: 1 to update the node_sets, 2 to print the warning
```suggestion
            overwritten = node_sets.update(NodeSets.from_file(path))
            _warn_on_overwritten_node_sets(overwritten)
```

bluepysnap/simulation.py
```diff
@@ -125,8 +136,17 @@ def simulator(self):
     @cached_property
     def node_sets(self):
         """Returns the NodeSets object bound to the simulation."""
-        path = self.to_libsonata.node_sets_file
-        return NodeSets.from_file(path) if path else NodeSets.from_dict({})
+        try:
+            path = self.circuit.to_libsonata.node_sets_path
+        except BluepySnapError:
```
GianlucaFicarelli, line 141: I see that `self.circuit` might raise BluepySnapError if the circuit config file doesn't exist.
Is the simulation config considered valid even in this case?
If we need to keep the check, a comment with the type of expected error may be useful.

bluepysnap/simulation.py
```diff
@@ -125,8 +136,17 @@ def simulator(self):
...
+            path = ""
+
+        node_sets = NodeSets.from_file(path) if path else NodeSets.from_dict({})
```
GianlucaFicarelli, line 144: We could get a copy of `self.circuit.node_sets` instead, but it makes sense only if it makes the code simpler.

bluepysnap/simulation.py
joni-herttuainen: Done

bluepysnap/simulation.py
joni-herttuainen: Not commenting on the validity of the simulation but, according to the spec, `network` is an optional entry, so I thought it'd be better to not require it being there for the node sets, either.

-> Added a comment

bluepysnap/simulation.py
joni-herttuainen: Unless there's another use case for copying a node set, I'm not sure it's worth it.

bluepysnap/node_sets.py
joni-herttuainen: Agreed -> done

bluepysnap/node_sets.py
```diff
@@ -67,7 +68,7 @@ def __init__(self, content, instance):
         Returns:
```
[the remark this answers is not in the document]
joni-herttuainen: -> done

GianlucaFicarelli (approved):
LGTM

PR #238, opened 2023-10-17
Fix `cyclic-import` issues

joni-herttuainen:
Also unrestrict `pylint` version.

### DIFF
--- a/bluepysnap/circuit_ids.py
+++ b/bluepysnap/circuit_ids.py
@@ -17,7 +17,6 @@
 
 """Circuit ids."""
 import abc
-from collections import namedtuple
 
 import numpy as np
 import pandas as pd
@@ -26,8 +25,8 @@ from bluepysnap import utils
 from bluepysnap._doctools import AbstractDocSubstitutionMeta
 from bluepysnap.exceptions import BluepySnapError
 
-CircuitNodeId = namedtuple("CircuitNodeId", ("population", "id"))
-CircuitEdgeId = namedtuple("CircuitEdgeId", ("population", "id"))
+CircuitNodeId = utils.CircuitNodeId
+CircuitEdgeId = utils.CircuitEdgeId
 
 
 class CircuitIds(abc.ABC):
--- a/bluepysnap/circuit_ids.py
+++ b/bluepysnap/circuit_ids.py
@@ -23,6 +23,7 @@ import pandas as pd
 
 from bluepysnap import utils
 from bluepysnap._doctools import AbstractDocSubstitutionMeta
+from bluepysnap.circuit_ids_types import CircuitEdgeId, CircuitNodeId
 from bluepysnap.exceptions import BluepySnapError
 
 
@@ -320,11 +321,11 @@ class CircuitNodeIds(
     def __iter__(self):
         """Iterator on the CircuitNodeIds."""
         for index in self.index:
-            yield utils.CircuitNodeId(*index)
+            yield CircuitNodeId(*index)
 
     def __getitem__(self, item):
         """Getter on the CircuitNodeIds."""
-        return utils.CircuitNodeId(*self.index[item])
+        return CircuitNodeId(*self.index[item])
 
 
 class CircuitEdgeIds(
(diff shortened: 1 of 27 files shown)


bluepysnap/circuit_validation.py
```diff
@@ -12,6 +12,7 @@
 
 from bluepysnap import schemas
 from bluepysnap.config import Parser
+from bluepysnap.exceptions import BluepySnapValidationError as Error
```
joni-herttuainen, line 15: I'm on the fence if I want to use the alias `Error` here. 

I just did it to minimize the changes to the file, as `BluepySnapValidationError` is verbose enough to have black restructure the code quite a bit.

Any thoughts?

bluepysnap/circuit_ids.py
```diff
@@ -324,11 +320,11 @@ def __init__(self, index, sort_index=True):
     def __iter__(self):
         """Iterator on the CircuitNodeIds."""
         for index in self.index:
-            yield CircuitNodeId(*index)
+            yield utils.CircuitNodeId(*index)
```
GianlucaFicarelli, line 323: I'm not fully convinced that `CircuitNodeId` belongs to `util`.
What if we move `CircuitNodeId`, `CircuitEdgeId` (and maybe also `IDS_DTYPE`) to a different file, for example `circuit_ids_types.py`?

bluepysnap/circuit_validation.py
GianlucaFicarelli: Same thoughts, what about using `as ValidationError`? Would the changes be smaller?

joni-herttuainen: I'll give it a go.

bluepysnap/circuit_ids.py
joni-herttuainen: > I'm not fully convinced that CircuitNodeId belongs to util.
Yeah, that was what I was contemplating too, but I didn't feel the functions belong to `circuit_ids.py` either nor did I want to introduce another file. 

In retrospect, I think another file is cleaner. I'll change it.

joni-herttuainen: Done

bluepysnap/circuit_validation.py
joni-herttuainen: Fixed: created a `warning` shortcut function to reduce the impact and kept the longer form to follow the style of the other exceptions.

GianlucaFicarelli: Nice, I like it more.

bluepysnap/schemas/schemas.py
```diff
@@ -79,10 +80,10 @@ def _wrap_errors(filepath, schema_errors, join_str):
 
     if len(warnings) > 0:
         message = filepath + ":\n\t" + "\n\t".join(warnings)
-        ret_errors.append(Error(Error.WARNING, message))
+        ret_errors.append(BluepySnapValidationError(BluepySnapValidationError.WARNING, message))
     if len(errors) > 0:
         message = filepath + ":\n\t" + "\n\t".join(errors)
-        ret_errors.append(Error(Error.FATAL, message))
+        ret_errors.append(BluepySnapValidationError(BluepySnapValidationError.FATAL, message))
```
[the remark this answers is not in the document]
joni-herttuainen: Sadly, no. Unless we move the functions somewhere else. Otherwise, we'd have cyclic import: 
`circuit_validation -> schemas -> circuit_validation`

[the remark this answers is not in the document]
joni-herttuainen: On another note, I think It'd be worth it to use the functions it in the `test_circuit_validation.py` to make it a bit cleaner.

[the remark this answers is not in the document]
joni-herttuainen: You know what. I'll add them to the BluepySnapValidationError:
```python
class BluepySnapValidationError:
...
    @classmethod
    def warning(cls, msg):
        return cls(cls.WARNING, msg)
...
```
and see where that takes us.

GianlucaFicarelli (approved):

bluepysnap/schemas/schemas.py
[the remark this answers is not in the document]
GianlucaFicarelli: Nice, it looks cleaner 👍

PR #240, opened 2023-10-20
Add very basic testing for `_plotting.py`, fix a few minor things

joni-herttuainen:
The intention is to cover all lines so that any deprecation warnings etc. don't go unnoticed. I.e., the testing does not ensure that correct things are plotted nor that things are plotted correctly.

### DIFF
--- a/bluepysnap/_plotting.py
+++ b/bluepysnap/_plotting.py
@@ -207,7 +207,9 @@ def spikes_isi(filtered_report, use_frequency=False, binsize=None, ax=None):
     if binsize is not None and binsize <= 0:
         raise BluepySnapError(f"Invalid binsize = {binsize}. Should be > 0.")
 
-    gb = filtered_report.report.groupby(["ids", "population"], observed=False)
+    # Added `observed=True` to silence pandas warning about changing default value.
+    # However, report should not contain categories that are not in the dataframe.
+    gb = filtered_report.report.groupby(["ids", "population"], observed=True)
     values = np.concatenate([np.diff(node_spikes.index.to_numpy()) for _, node_spikes in gb])
 
     if len(values) == 0:
--- a/bluepysnap/_plotting.py
+++ b/bluepysnap/_plotting.py
@@ -40,7 +40,7 @@ def _get_pyplot():
     return plt
 
 
-def spikes_firing_rate_histogram(filtered_report, time_binsize=None, ax=None):  # pragma: no cover
+def spikes_firing_rate_histogram(filtered_report, time_binsize=None, ax=None):
     """Spike firing rate histogram.
 
     This plot shows the number of nodes firing during a range of time.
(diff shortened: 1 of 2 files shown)


bluepysnap/_plotting.py
```diff
@@ -204,7 +207,7 @@ def spikes_isi(filtered_report, use_frequency=False, binsize=None, ax=None):  #
     if binsize is not None and binsize <= 0:
         raise BluepySnapError(f"Invalid binsize = {binsize}. Should be > 0.")
 
-    gb = filtered_report.report.groupby(["ids", "population"])
+    gb = filtered_report.report.groupby(["ids", "population"], observed=False)
```
joni-herttuainen, line 210: `observed=False` is to keep the current behavior and to silence the pandas warning that it will default to `observed=True` in the future.

However, I'm not convinced we should keep observed=False.

GianlucaFicarelli: From what I can see, it doesn't make any difference when `filtered_report` is a spike report, because `filtered_report.report` returns a dataframe with columns `ids` and `population`, where only `population` is categorical, and it's explicitly converted to category [here](https://github.com/BlueBrain/snap/blob/6315a6b8299c4d5396c84fbfb84c146b4c5b7b86/bluepysnap/spike_report.py#L175)

So it's not possible to have categories not observed in the actual data.
I think we can set `observed=True`, that will be the default in a future version, and maybe add a short comment to explain the reason.

tests/test__plotting.py
```diff
@@ -15,3 +25,140 @@ def test__get_pyplot():
...
+def get_report(dlen=100):
+    rng = np.random.default_rng(seed=42)
+    return pd.DataFrame(
+        {
+            "population": dlen * ["default"],
+            "property": pd.Series(dlen * ["value"], dtype="category"),
+            "ids": rng.integers(1, 100, dlen),
+        },
+        index=rng.integers(1, 100, dlen),
+    )
```
GianlucaFicarelli, line 39: The mocked (spike) report should be like
```
       ids population
times                
0.1      2    default
0.1      2   default2
0.2      0   default2
```
where:
- the population column should be: categorical
- ids: int64
- times: float64 (as index)

bluepysnap/_plotting.py
GianlucaFicarelli: From what I can see, it doesn't make difference when `filtered_report` is a spike report, because `filtered_report.report` returns a dataframe with columns `ids` and `population`, where only `population` is categorical, and it's explicitly converted to category [here](https://github.com/BlueBrain/snap/blob/6315a6b8299c4d5396c84fbfb84c146b4c5b7b86/bluepysnap/spike_report.py#L175)

So it's not possible to have categories not observed in the actual data.
I think we can set `observed=True`, that will be the default in a future version, and maybe add a short comment to explain the reason.

tests/test__plotting.py
joni-herttuainen: Yeah, this was mostly leftovers, I removed it and just `patch`ed `np.concatenate` to return empty array instead.

GianlucaFicarelli (approved):

PR #241, opened 2023-10-23
Have reports use Simulation Node Sets

joni-herttuainen:
@GianlucaFicarelli : please have a look and let me know what you think of the approach.

I might've gotten tunnel-visioned here, and maybe there's a simpler approach I failed to see. 

I can add more tests and explicit tests for `query.resolve_nodesets` but I'll rather put extra effort before you take a look and say that this is the way to go. 😅


GianlucaFicarelli (commented):
The approach looks good to me 👍

mgeplf (approved):

PR #247, opened 2023-11-23
Add further simulation validation

joni-herttuainen:
Have simulation validation check 
* the existence of the files and node sets defined in the configuration
* that the node ids and populations defined in input spike files are found in the corresponding nodesets and populations
* that the node ids and populations defined in ``electrodes_file`` are found in the corresponding nodesets and populations
* mechanisms and modoverrides are correct (if `neurodamus` found in the environment)

**NOTE:** Ready for review, but let's not merge before @mgeplf has had the chance to check if we can do the mechanisms and modoverrides check without involving `neurodamus`


mgeplf (approved):

joni-herttuainen:
I'll push this the current `.dat` still supported version and open a new PR for removing the support and change it according to whatever we decide in https://github.com/BlueBrain/sonata-extension/pull/43

PR #250, opened 2023-11-29
Add include_empty keyword argument

HDictus:
When using get-like functions (edges.pathway_edges, nodes.get) it is common to concatenate the outputs of the functions to mimic the old bluepysnap interface. However, if no nodes/edges satisfy the query, instead of resulting in an empty dataframe, this approach will result in an error, because no dataframes at all are yielded by the generator.

The include_empty keyword argument allows the user to specify that they still want to see a dataframe if the query turns up nothing.

One example where this is useful for me is when I use the length of a connection dataframe to determine connection probability.

### DIFF
--- a/bluepysnap/edges/edges.py
+++ b/bluepysnap/edges/edges.py
@@ -82,7 +82,11 @@ class Edges(
         fun = lambda x: (x.ids(group), x.name)
         return self._get_ids_from_pop(fun, CircuitEdgeIds, sample=sample, limit=limit)
 
-    def get(self, edge_ids=None, properties=None):  # pylint: disable=arguments-renamed
+    def get(self, 
+            edge_ids=None, 
+            properties=None, 
+            include_empty=False
+        ):  # pylint: disable=arguments-renamed
         """Edge properties by iterating populations.
 
         Args:
@@ -102,7 +106,7 @@ class Edges(
             raise BluepySnapError("You need to set edge_ids in get.")
         if properties is None:
             return edge_ids
-        return super().get(edge_ids, properties)
+        return super().get(edge_ids, properties, include_empty=include_empty)
 
     def afferent_nodes(self, target, unique=True):
         """Get afferent CircuitNodeIDs for given target ``node_id``.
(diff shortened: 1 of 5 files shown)


mgeplf:
Thanks for the PR.

>  `it is common to concatenate the outputs of the functions to mimic the old bluepysnap interface`

I'm generally a bit leary to add more arguments to functions, since it's a bit a of a slipperly slope (but I'm not ruling this out).
Out of curiousity, is it not possible to do something like this to filter the empty ones:
```
dfs = (df for df in foo.get({"layer": 10}) if not df.empty)
```

HDictus:
> 
> ```
> dfs = (df for df in foo.get({"layer": 10}) if not df.empty)
> ```

The current behavior does not yield anything if the query comes up empty. So `foo.get({"layer": 10})` would give an empty iterator. If .get instead yielded empty dataframes then I would not need the keyword argument. However, that would be a potentially breaking change, which is why I've put a kwarg and default = False. Having the iterator include empty dataframes without a kwarg would be preferable for me as well, if you think it would be ok to change the behavior that way I would be happy to change the PR to that.

joni-herttuainen:
If I understood correctly, the use case is to `pd.concat` the dataframes. If this is the use case where you have a wrapper for the dataframes, you can do something like:

```python
def _concat(iterator):
    dfs = [df for _,df in iterator]
    return pd.concat(dfs) if dfs else pd.DataFrame([])  # or if you just want to check the length, you might just have "else []"

df = _concat(c.nodes.get(query))
```
or, you can use `dict(...).values()`:
```python
def _concat(iterator):
    dfs = dict(iterator).values() 
    return pd.concat(dfs) if dfs else pd.DataFrame([])  # or "else []" if just for length
```
You can ever one-line it, albeit the readability takes a hit:
```python
def _concat(iterator):
     return pd.concat(d.values()) if (d := dict(iterator)) else pd.DataFrame([])
```

HDictus:
The issue with this is that downstream code which expects the dataframe to have columns will break because it has none, and so I will need to add a if df.empty then do something else clause to everything.

HDictus:
I often only retrieve a subset of the columns, so then I'll need to pass this subset to the function as well. In this case I may be best off wrapping the .get method call as a whole
```
def get_dataframe(networkobj, *args, properties=None, attr='get'):
     dfs = [df for _ df in getattr(networkobj, attr)(*args, properties=properties)]
     if len(dfs) == 0:
        return pd.DataFrame(columns=properties or networkobj.property_names)
     return pd.concat(dfs)
```

So strictly speaking it isn't necessary to have the empty dataframes, just simpler and more concise.

PR #251, opened 2023-12-05
Issue reading LFP reports

jorblancoa:
Addresses https://bbpteam.epfl.ch/project/issues/browse/NSETM-2269

Change the logic to also check the report "type" and if it is "lfp" select a CompartmentReport

### DIFF
--- a/bluepysnap/simulation.py
+++ b/bluepysnap/simulation.py
@@ -32,17 +32,22 @@ def _collect_frame_reports(sim):
     res = {}
     for name in sim.to_libsonata.list_report_names:
         report = sim.to_libsonata.report(name)
-        report_type = report.sections.name
-        if report_type == "soma":
-            from bluepysnap.frame_report import SomaReport
-
-            cls = SomaReport
-        elif report_type == "all":
+        report_type = report.type.name
+        if report_type == "lfp":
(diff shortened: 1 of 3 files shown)


bluepysnap/simulation.py
```diff
@@ -32,17 +32,22 @@ def _collect_frame_reports(sim):
...
+            if report_sections == "soma":
+                from bluepysnap.frame_report import SomaReport
+
+                cls = SomaReport
+            elif report_sections == "all":
+                from bluepysnap.frame_report import CompartmentReport
+
+                cls = CompartmentReport
+            else:
+                raise BluepySnapError(f"Report {name}: format {report_sections} not yet supported.")
```
joni-herttuainen, line 50: I think this would be a bit cleaner
```python
        report_type = report.sections.name
        if report_type == "all" or report.type.name == "lfp":
            from bluepysnap.frame_report import CompartmentReport

            cls = CompartmentReport
        elif report_type == "soma":
            from bluepysnap.frame_report import SomaReport

            cls = SomaReport
        else:
            raise BluepySnapError(f"Report {name}: format {report_type} not yet supported.")

        res[name] = cls(sim, name)
    return res
```

joni-herttuainen (approved):
LGTM

PR #253, opened 2024-01-30
3.0.0 release

edasubert:


tests/test_schemas.py
```diff
@@ -187,8 +187,8 @@ def test_validate_config_ok_missing_optional_fields(to_remove):
             [["networks", "edges", 0, "populations"]],
             "networks.edges[0]: 'populations' is a required property",
         ),
-        ([["networks", "nodes", 0]], "networks.nodes: [] is too short"),
-        ([["networks", "edges", 0]], "networks.edges: [] is too short"),
+        ([["networks", "nodes", 0]], "networks.nodes: [] should be non-empty"),
+        ([["networks", "edges", 0]], "networks.edges: [] should be non-empty"),
```
edasubert, line 191: I am not sure about this, it fails with current master

joni-herttuainen: Yeah, seems this was changed in `jsonschema=4.21.0`: https://github.com/python-jsonschema/jsonschema/commit/8875c2ecb1094d8ba265a071d4c1ce9df1a94733 

Let's use the `should be non-empty`. I don't think there'll be many of these changes in the future, but if it happens, let's pin a version in the tox config. I.e., not now, but if these keep changing in the future.

CHANGELOG.rst
```diff
@@ -20,10 +12,10 @@ New Features
 
   - deprecated ``validate``
```
joni-herttuainen, line 13: these lines should also be changed

joni-herttuainen (approved):

PR #257, opened 2024-02-09
handle other iterables than list in queries

### LINKED ISSUE #256: `get` and `ids` functions do not work with tuples
Did not test with `Edges`/`EdgePopulation` but I would assume the issue is the same as with `Nodes`/`NodePopulation`:

```python
>>> circuit.nodes['sscx_neurons'].ids({'region': ['mc2_Column', 'mc3_Column']})
array([     0,      2,      7, ..., 163519, 163521, 163523])

>>> circuit.nodes['sscx_neurons'].ids({'region': ('mc2_Column', 'mc3_Column')})
array([], dtype=int64)
```

I am unsure if we explicitly specify that the input for `OR` needs to be given as a `list` but I'd expect it to work with other iterables, too.

joni-herttuainen:
Fixes #256

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Version v3.0.1
+--------------
+
+Bug Fixes
+~~~~~~~~~
+- Fixed a bug causing some iterables (e.g., tuples) in queries not to work as expected
+
+
 Version v3.0.0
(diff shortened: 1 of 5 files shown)


GianlucaFicarelli (approved):

mgeplf (approved):
Makes sense to me.

PR #259, opened 2024-02-14
Combine `sonata_constants.py` and `bbp.py`

joni-herttuainen:
* Combine Node and Cell, keep aliases
 * Combine Edge and Synapse, keep aliases
 * deprecate bbp.py

### DIFF
--- a/bluepysnap/bbp.py
+++ b/bluepysnap/bbp.py
@@ -17,45 +17,21 @@
 
 """BBP cell / synapse attribute namespace."""
 
-from bluepysnap.sonata_constants import DYNAMICS_PREFIX, Edge, Node
+import warnings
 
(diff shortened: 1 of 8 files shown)


bluepysnap/bbp.py
```diff
@@ -17,45 +17,21 @@
 
 """BBP cell / synapse attribute namespace."""
 
-from bluepysnap.sonata_constants import DYNAMICS_PREFIX, Edge, Node
```
mgeplf, line 20: after giving this more thought, I think it might be worthwhile to just keep the BBP contants for now; it makes porting easier for internal code, but putting them into the main namespace and having more constants I think makes it more confusing.

What do you think?

joni-herttuainen: Sure, makes sense. 🤘

PR #264, opened 2024-03-11
Create aliases for `{a,e}fferent_{edges,nodes}`

joni-herttuainen:
Oftentimes, the naming of `afferent_edges`/`efferent_edges`/`afferent_nodes`/`efferent_nodes` cause confusion among our users. Also, quite frankly, I always have to look them up too as they don't feel intuitive.  It's not clear what they mean and when is it needed to query by source nodes and when by target nodes.

To remove the ambiguity without introducing breaking changes, I created aliases for them (naming is open for discussion):
```python
edges_by_source = efferent_edges
edges_by_target = afferent_edges
target_nodes_by_source = efferent_nodes
source_nodes_by_target = afferent_nodes
```

### DIFF
--- a/bluepysnap/edges/edge_population.py
+++ b/bluepysnap/edges/edge_population.py
@@ -627,6 +627,12 @@ class EdgePopulation:
             raise BluepySnapError(f"It appears {self.name} does not have synapse indices")
         return open_index(index_dir)
 
+    # ALIASES
+    edges_by_source = efferent_edges
+    edges_by_target = afferent_edges
+    target_nodes_by_source = efferent_nodes
+    source_nodes_by_target = afferent_nodes
+
     def __getstate__(self):
         """Make EdgePopulation pickle-able, without storing state of caches."""
         return self._circuit, self.name
--- a/bluepysnap/edges/edges.py
+++ b/bluepysnap/edges/edges.py
@@ -250,6 +250,12 @@ class Edges(
                 return_edge_count=return_edge_count,
             )
 
+    # ALIASES
+    edges_by_source = efferent_edges
+    edges_by_target = afferent_edges
+    target_nodes_by_source = efferent_nodes
+    source_nodes_by_target = afferent_nodes
+
     def __getstate__(self):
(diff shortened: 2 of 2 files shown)


bluepysnap/edges/edge_population.py
```diff
@@ -627,6 +627,12 @@ def spatial_synapse_index(self):
             raise BluepySnapError(f"It appears {self.name} does not have synapse indices")
         return open_index(index_dir)
 
+    # ALIASES
+    edges_by_source = efferent_edges
+    edges_by_target = afferent_edges
+    target_nodes_by_source = efferent_nodes
+    source_nodes_by_target = afferent_nodes
```
matz-e, line 634: I find the naming a bit clunky. `outgoing_edges` and `incoming_edges` would be IMO more intuitive for eff and aff edges.

`target_nodes_by_source` also sounds very confusing!

joni-herttuainen: > I find the naming a bit clunky. `outgoing_edges` and `incoming_edges` would be IMO more intuitive for eff and aff edges.

Sure, whatever makes sense to everybody is fine by me.

> `target_nodes_by_source` also sounds very confusing!

What would you suggest? I can always elaborate it with `get_target_nodes_of_source_nodes` or whatever. I just want something that people will immediately understand so they don't accidentally choose the incorrect function for their purposes.

GianlucaFicarelli: What about just `target_nodes` and `source_nodes`?
They would match the names of the underlying libsonata edges population [target_nodes](https://libsonata.readthedocs.io/en/stable/api.html#libsonata.EdgePopulation.target_nodes) and [source_nodes](https://libsonata.readthedocs.io/en/stable/api.html#libsonata.EdgePopulation.source_nodes)

joni-herttuainen: In libsonata, they expect edge ids as input (as opposed to node ids), right? Wouldn't that be even more confusing?

PR #265, opened 2024-04-05
add option to ignore datatypes while validating

mgeplf:
* teach the `bluepysnap validate-circuit` and `bluepysnap validate-simulation` the ability to `--ignore-datatype-errors` so that mismatches of datatypes to the specification are ignored

### DIFF
--- a/bluepysnap/cli.py
+++ b/bluepysnap/cli.py
@@ -1,6 +1,5 @@
 """The project's command line launcher."""
 
-import functools
 import logging
 
 import click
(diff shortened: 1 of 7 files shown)


bluepysnap/cli.py
```diff
@@ -22,45 +21,48 @@ def cli(verbose):
     )
 
 
-def circuit_validation_params(func):
```
joni-herttuainen, line 25: I think I did this to have the option to validate the circuit in the simulation config but opted out of implementing the feature as it seemed overly complicated. 🤔 

Anyway, happy to remove this leftover.

joni-herttuainen (approved):
LGTM

PR #267, opened 2024-07-08
Fix tests for `numpy>=2`

joni-herttuainen:


bluepysnap/circuit_validation.py
```diff
@@ -20,6 +20,10 @@
 L = logging.getLogger("brainbuilder")
 MAX_MISSING_FILES_DISPLAY = 10
 
+# print numpy scalars as "1" instead of "np.uint64(1)" (changed in numpy==2.0.0)
+if np.version.version >= "2.0.0":  # pragma: no cover
+    np.set_printoptions(legacy="1.25")
```
mgeplf, line 25: hmm, could this be done more localized?
setting a global setting at import gives me the heebie jeebies.

joni-herttuainen: Yeah, no worries, I see what I can do.

On a side note: I like the term, "heebie jeebies" 😆

mgeplf (approved):

PR #268, opened 2024-08-30
Fix: stats module use CircuitNodeId

arnaudon:


joni-herttuainen:
@GianlucaFicarelli : Can you review this?

PR #269, opened 2024-09-02
Fix missing units so tests run again

mgeplf:


tests/data/reporting/create_reports.py
```diff
@@ -16,7 +18,8 @@ def write_spikes(filepath):
         gpop_spikes = h5f.create_group("/spikes/" + population_names[0])
         gpop_spikes.attrs.create("sorting", data=2, dtype=sorting_type)
         timestamps, node_ids = zip(*sorted(zip(timestamps_base, node_ids_base)))
-        gpop_spikes.create_dataset("timestamps", data=timestamps, dtype=np.double)
+        dtimestamps = gpop_spikes.create_dataset("timestamps", data=timestamps, dtype=np.double)
+        dtimestamps.attrs.create("units", data="ms", dtype=string_dtype)
```
GianlucaFicarelli, line 22: If I'm not wrong units should be added also to the second population default2 (gpop_spikes2)

GianlucaFicarelli (approved):

PR #272, opened 2024-10-09
Switch to the open-sourced brain-indexer

matz-e:
Remove the distutils dependency in the test setup, too.  This module
will not be present any longer in Python 3.12.

Removed tests are duplicated and fail, as spatial indices are only
defined for the second populations of edges and nodes.

### DIFF
--- a/tests/test_edge_population.py
+++ b/tests/test_edge_population.py
@@ -728,27 +728,10 @@ class TestEdgePopulation:
     def test_h5_filepath_from_config(self):
         assert self.test_obj.h5_filepath == str(TEST_DATA_DIR / "edges.h5")
 
-    @pytest.mark.skip(reason="Until spatial-index is released publicly")
-    def test_spatial_synapse_index(self):
-        with mock.patch("spatial_index.open_index") as mock_open_index:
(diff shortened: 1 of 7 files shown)


mgeplf:
let me update the python versions

heerener (approved):

tests/test_edge_population.py
```diff
@@ -728,30 +728,13 @@ def test_iter_connection_unique(self):
     def test_h5_filepath_from_config(self):
         assert self.test_obj.h5_filepath == str(TEST_DATA_DIR / "edges.h5")
 
-    @pytest.mark.skip(reason="Until spatial-index is released publicly")
-    def test_spatial_synapse_index(self):
```
mgeplf, line 732: why delete this one?

matz-e: This is the first population that does not define the spatial index directory.

Besides, at the end of the file, this test is done for the second edge population with a defined spatial index path, and done properly.

(the call will contain an absolute path, and not just the final fragment)

mgeplf (approved):

mgeplf:
Thanks!

PR #274, opened 2024-10-30
Support morphology containers

mgeplf:
* use the `morphio.Container`, to both load from a directory or from a file


CHANGELOG.rst
```diff
@@ -20,6 +20,7 @@ Improvements
 - Update simulation validation to conform to the SONATA spec
 
   - ``synapse_replay.source`` and ``.dat`` spike input files are no longer supported
+- Support morphology containers
```
GianlucaFicarelli, line 23: It would be needed to update the minimum version of morphio in setup.py (it's still >=3.0.0, but it seems that container support has been added later)

bluepysnap/morph.py
```diff
@@ -90,11 +107,18 @@ def get(self, node_id, transform=False, extension="swc"):
...
         """
-        filepath = self.get_filepath(node_id, extension=extension)
-        result = Morphology(filepath)
+        if extension == "h5":
+            collection = morphio.Collection(self._get_morphology_base(extension))
+            name = self.get_name(node_id)
+            result = collection.load(name, mutable=True)
+        else:
+            filepath = self.get_filepath(node_id, extension=extension)
+            result = morphio.mut.Morphology(filepath)
```
GianlucaFicarelli, line 116: I see that `morphio.Collection` can be used to load morphologies in both cases (when the collection is a h5 file, and when it's a directory containing multiple files).
Would the code be simplified if we always use the collection, instead of calling morphio.mut.Morphology?

matz-e (approved):
Thank you!

bluepysnap/morph.py
matz-e: Yes, that would be a possibility. Also, one could factor out the collection constructions and buffer it with `@lru_cache` or so...

CHANGELOG.rst
mgeplf: yeah, good point, will do

bluepysnap/morph.py
mgeplf: It's strange, I thought I was already doing that; so I'm not sure why this is an older one?

Let me look through my history.

> Also, one could factor out the collection constructions and buffer it with @lru_cache

is collection creation that expensive?
```
In [2]: %time morphio.Collection('/home/gevaert/src/snap/tests/data/morphologies')
CPU times: user 33 μs, sys: 2 μs, total: 35 μs
Wall time: 36 μs
Out[2]: <morphio._morphio.Collection at 0x7f8c5e9a4f30>
```

matz-e: It would save opening the file and traversing the file system tree to do so. Plus whatever internal accounting HDF5 does. May matter when many thousand processes do it at the same time.

mgeplf: I think we can cross that bridge when we get proof that it's a problem; if people are opening morphology files, the bottleneck is loading them, compared to having a Collection, imo.

bluepysnap/morph.py
```diff
@@ -68,16 +68,33 @@ def get_morphology_dir(self, extension):
...
         return morph_dir
 
+    def get_morphology_dir(self, extension="swc"):
+        """Return morphology directory based on a given extension."""
+        morph_dir = self._get_morphology_base(extension)
+
+        if extension == "h5" and Path(morph_dir).is_file():
+            raise BluepySnapError(
+                f"'{morph_dir}' is a morphology container, so a directory does not exist"
+            )
```
GianlucaFicarelli, line 78: If the circuit is using a h5 container instead of a directory, I wonder if we want to raise an error anytime that `get_morphology_dir` or `get_filepath` are called, regardless of the extension specified as a parameter?
Probably in the medium/long term those functions could be deprecated, but need to check in what use cases they are needed at the moment.

tests/test_morph.py
```diff
@@ -126,6 +126,25 @@ def test_get_morphology(self):
...
+            None, self.nodes, alternate_morphologies=alternate_morphs
+        )
+
+        node_id = 0
+
+        with pytest.raises(BluepySnapError):
+            test_obj.get_morphology_dir(extension="h5")
+
+        with pytest.raises(BluepySnapError):
+            test_obj.get_filepath(node_id, extension="h5")
```
GianlucaFicarelli, line 143: Minor, but the specific reason of BluepySnapError could be checked with `match=...`

bluepysnap/morph.py
mgeplf: You mean if it's the only `alternaive-morphology`?

GianlucaFicarelli: Yeah, but after checking better it's already good as is, because the container file is retrieved only with extension=h5

